PR #3, opened 2019-06-20
Fix: valid openapi definitions

ioggstream:
## This PR

Uses valid OAS definitions:

- add missing `title`
- add paths object
- remove invalid `$ref` in `definitions`

Tests works locally with circleci build. 

validator.swagger.io badge

Checks should be done on depending projects.


ioggstream:
In admin-api.yaml using an "old-style but invalid" reference works.

```
definitions:
  $ref: "https://raw.githubusercontent.com/teamdigitale/io-functions-commons/3c1ce27905798f87bdae08f46697edc38bb4ca99/openapi/index.yaml#/definitions"
```

cloudify:
@ioggstream can you check PR #16 ?

ioggstream:
@cloudify see comments there :)

PR #31, opened 2019-10-10
[#168821506] Functions and types required by the email validation process

francescopersico:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Please include a description of your PR changes.</td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#168821506](https://www.pivotaltracker.com/story/show/168821506): Quando nel profilo viene cambiata l'email deve partire il processo di verifica

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 9e1f7f33779728510a29c62efd1adcf6a9930421
</p>

openapi/definitions.yaml
```diff
@@ -322,14 +322,42 @@ BlockedInboxOrChannels:
   description: |-
     All the notification channels blocked by the user.
     Each channel is related to a specific service (sender).
+NewProfile:
```
cloudify: why is this needed? if you add `is_email_validated` to `Profile` you should get the same result

openapi/definitions.yaml
```diff
@@ -322,14 +322,42 @@ BlockedInboxOrChannels:
...
+    - is_email_validated
 Profile:
+  type: object
   description: |-
-    Describes the citizen's profile, mostly interesting for preferences
-    attributes.
-  x-one-of: true
-  allOf:
-    - $ref: "#/LimitedProfile"
-    - $ref: "#/ExtendedProfile"
```
cloudify: is `Profile` the same as `ExtendedProfile` ?

src/entities/verification_token.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import * as t from "io-ts";
+
+import { UTCISODateFromString } from "italia-ts-commons/lib/dates";
+import { FiscalCode } from "../../generated/definitions/FiscalCode";
+import { TableEntity } from "../utils/azure_storage";
+
+export const VerificationToken = t.interface({
+  ExpireAt: UTCISODateFromString,
```
cloudify: more than an event in time that may be confusing, it's better to describe a portion of time, like `validBefore` or `invalidAfter` (depends whether you include that specific time in the interval)

also we use camel case for attribute names

src/utils/azure_storage.ts
```diff
@@ -8,14 +8,15 @@ import { Either, left, right } from "fp-ts/lib/Either";
...
 
-export const BlobNotFoundCode = "BlobNotFound";
-
 type Resolve<T> = (value?: T | PromiseLike<T>) => void;
 
 type StorageError = Error & {
   code?: string;
 };
 
+// Blob storage stuff
```
cloudify: needs a better comment :)

src/utils/azure_storage.ts
```diff
@@ -222,3 +223,106 @@ export async function getBlobAsObject<A, O, I>(
     }
   });
 }
+
+// Table storage stuff
```
cloudify: same

src/utils/azure_storage.ts
```diff
@@ -222,3 +223,106 @@ export async function getBlobAsObject<A, O, I>(
...
+ * `"Key": {"&": DataType, "_": Value}`
+ *
+ * this function converts to an object containing only the values like:
+ *
+ * `"Key": Value`
+ *
+ * @param entityResult the object returned by TableService retrieveEntity function
+ */
+// tslint:disable-next-line: no-any
+export function getValueOnlyEntityResolver(entityResult: any): any {
```
cloudify: use `unknown` instead of `any`

src/utils/azure_storage.ts
```diff
@@ -222,3 +223,106 @@ export async function getBlobAsObject<A, O, I>(
...
+  // tslint:disable-next-line: no-any
+  return Object.keys(entityResult).reduce<any>((accumulator, key) => {
```
cloudify: this will thrown an exception if `entityResult` is not an object

src/utils/azure_storage.ts
```diff
@@ -222,3 +223,106 @@ export async function getBlobAsObject<A, O, I>(
...
+ */
+export async function insertTableEntity<T extends ITableEntity>(
+  tableService: azureStorage.TableService,
+  tableName: string,
+  entity: T
+): Promise<Either<Error, azureStorage.TableService.EntityMetadata>> {
+  return new Promise(resolve => {
+    // tslint:disable-next-line: no-identical-functions
+    tableService.insertEntity<T>(tableName, entity, (err, result, _) => {
+      if (err) {
```
cloudify:
```ts
return resolve(err ? left(err) : right(result))
```

src/utils/azure_storage.ts
```diff
@@ -222,3 +223,106 @@ export async function getBlobAsObject<A, O, I>(
...
+          if (errorAsStorageError.code === ResourceNotFoundCode) {
+            return resolve(right(none));
+          }
+          return resolve(left(err));
+        } else {
+          return resolve(
+            type
+              .decode(result)
+              .fold(
+                e => left(new Error(readableReport(e))),
```
cloudify: keep the io-ts `Errors ` type here, we usually translate to a string with `readableReport` just before rendering the error (e.g. in logs)

openapi/definitions.yaml
francescopersico: @cloudify 
Let me try to explain the logic of the types:

**NewProfile**: contains is_email_validated, this field is needed because the Backend on first user login CREATE the profile and set is_email_validated to true if the login is done with SPID and to false otherwise.

**Profile**: do NOT contains is_email_validated because i don't want the Backend nor the App to directly manipulate this field, only the verification process can change it. An alternative is to accept it in the type and ignore it on update but i think it is a violation of the "contract" established through the specification.

**ExtendedProfile**: contains is_email_validated because we want to return ALL the profile information to the Backend and than the App.

openapi/definitions.yaml
francescopersico: Read above

src/entities/verification_token.ts
francescopersico: Camel case -> Are you sure you want a table to have "PartitionKey, RowKey, Version" and than "myCamelCaseProperty"? All the system tables (like the Hub* tables created by the functions) use property in PascalCase.

ExpireAt -> I'm not sure I understood what you mean. Are you just suggesting to rename ExpireAt to invalidAfter/validBefore?

openapi/definitions.yaml
cloudify: ok makes sense

src/entities/verification_token.ts
cloudify: re case -> can you think of a way to make it clear these types are used to table props and because of that they have a different naming convention? perhaps a comment on top of this file? another reader may have the same concert I had

expire at -> yes. the "at" of expireAt is not clear about the `=` part of the comparison, is it still valid or invalid when `t == expireAt` ? on the other hand if you define it with `before` or `after` it's clear whether it's valid for `t < expireAt` (valid before) or `t <= expireAt` (invalid after)

src/utils/azure_storage.ts
```diff
@@ -317,7 +318,7 @@ export async function retrieveTableEntity<A, O, I>(
   options: azureStorage.TableService.TableEntityRequestOptions = {
     entityResolver: getValueOnlyEntityResolver
   }
-): Promise<Either<Error, Option<A>>> {
+): Promise<Either<Error | ReadonlyArray<t.ValidationError>, Option<A>>> {
```
cloudify: interpreting `Error | ReadonlyArray<t.ValidationError>` will require runtime checks - we should stick to either one

one option would be that, since `Error` is not well typed anyway, to just reject the promise with the error in case of non-404 errors

another option would be to return a nested `Either`:  `Promise<Either<Either<Error, ReadonlyArray<t.ValidationError>>, Option<A>>>`

however, when return types become so complicated, it's usually a sign that the function is trying to do too many things - it could be made simpler and rely on composing multiple functions - if we remove the entity decoding from this function, we can have it return a:

`Promise<Either<StorageError, Option<unknown>>>`

then pipe the result into the io-ts decode with a map

francescopersico: Done

gunzip:
@francescopersico can we cloe this one ? it looks like the code is already into master

gunzip:
closing

PR #42, opened 2020-02-10
[#170127598] extended service model

alexgpeppe:
This PR aims to extend the `Service` model  by adding the metadata defined in [io-services-metadata](https://github.com/teamdigitale/io-services-metadata/blob/master/definitions.yml#L6).

### DIFF
--- a/src/models/service.ts
+++ b/src/models/service.ts
@@ -1,5 +1,6 @@
 import * as t from "io-ts";
 import { PathReporter } from "io-ts/lib/PathReporter";
+import { enumType } from "italia-ts-commons/lib/types";
 
 import * as DocumentDb from "documentdb";
 import * as DocumentDbUtils from "../utils/documentdb";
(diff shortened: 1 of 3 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#170127598](https://www.pivotaltracker.com/story/show/170127598): Come backend voglio poter salvare le informazioni sui servizi in modo che possano essere utilizzate da IO

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against aad0e00c4444e6cd1e25a1ced3c9d318850a278e
</p>

src/models/service.ts
```diff
@@ -93,9 +67,7 @@ const ServiceMetadataO = t.partial({
 
   email: t.string,
 
-  pec: t.string,
-  // The attribute `scope` should be required, but it's currently optional for backward-compatibility reasons
-  scope: enumType<ScopeEnum>(ScopeEnum, "scope")
+  pec: t.string
```
gunzip: may you change all `t.string` into `NonEmptyString` ? (since the field is not required we can omit it in case it's e)

src/models/service.ts
```diff
@@ -62,6 +62,54 @@ export const Service = t.interface({
...
+
+export const ServiceMetadata = t.intersection(
+  [ServiceMetadataR, ServiceMetadataO],
+  "ServiceMetadata"
+);
+
+export type ServiceMetadata = t.TypeOf<typeof ServiceMetadata>;
+
+export const Service = t.intersection(
+  [ServiceBase, ServiceMetadata],
```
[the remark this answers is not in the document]
Undermaken: @gunzip  imho the second one could be better
In this way we can easily check if metadata field is defined or not.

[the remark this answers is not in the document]
gunzip: I agree :)

src/models/service.ts
```diff
@@ -35,10 +36,51 @@ import { MaxAllowedPaymentAmount } from "../../generated/definitions/MaxAllowedP
...
+// required attributes
+const ServiceMetadataR = t.interface({
+  scope: enumType<ScopeEnum>(ScopeEnum, "scope")
+});
+
+// optional attributes
+const ServiceMetadataO = t.partial({
+  description: NonEmptyString,
+
+  webUrl: NonEmptyString,
```
Undermaken: @gunzip I guess we prefer the snake_case, is it so?
 Indeed we defined that fields in snake_case in [io-service-metadata](https://github.com/pagopa/io-services-metadata/blob/master/definitions/Service.ts)

gunzip: this is the internal model used by Typescript and CosmosDB. The conversion in snake case happens here:https://github.com/pagopa/io-functions-admin/blob/master/utils/conversions.ts

PR #59, opened 2020-03-30
[#172050222] add prescription metadata into message model

gunzip:


openapi/definitions.yaml
```diff
@@ -106,6 +108,29 @@ NewMessage:
       $ref: "#/FiscalCode"
   required:
     - content
+PrescriptionData:
+  type: object
+  description: |-
+    Metadata needed to process medical prescriptions.
+  properties:
+    nre:
+      $ref: "#/PrescriptionNre"
```
cloudify, line 117: I believe `Nre` is an acronym, it should be all caps

openapi/definitions.yaml
```diff
@@ -106,6 +108,29 @@ NewMessage:
...
+    iup:
+      $ref: "#/PrescriptionIup"
```
cloudify, line 119: I believe `Iup` is an acronym, it should be all caps

openapi/definitions.yaml
```diff
@@ -106,6 +108,29 @@ NewMessage:
...
+    doctor_fiscal_code:
```
cloudify, line 120: in English, it's called `md` (Medicinae Doctor), so it should be `md_fiscal_code`, or perhaps if we want to be a little more generic we can use `prescriber_fiscal_code`.

gunzip: I'm ok with this but since these specs are already under review let's double check with @matteodesanti

gunzip:
closing in favor of #64

PR #67, opened 2020-04-21
Use fetch for making API calls to MailUP

cloudify:
need to fix tests


src/utils/mailup.ts
```diff
@@ -218,6 +227,8 @@ function toMailupAddress(
 export function MailUpTransport(
   options: IMailUpTransportOptions
 ): nodemailer.Transport {
+  const fetchAgent =
+    options.fetchAgent !== undefined ? options.fetchAgent : fetch;
```
gunzip, line 231: what's 'fetch' in this context, node-fetch ? (is it imported ?)

cloudify: you are right, I think it's just exported by `lib.dom`, not real

src/utils/mailup.ts
```diff
@@ -109,20 +106,29 @@ function callMailUpApi(
...
+    body: JSON.stringify({ ...payload, User: creds }),
+    method
+  })
     .then(response => {
-      if (response.error) {
+      if (!response.ok) {
         return left<Error, ApiResponse>(
           new Error(
-            `Error calling MailUp API: ${response.status} - ${response.text}`
+            `Error calling MailUp API: ${response.status} - ${response.text()}`
```
gunzip, line 120: response.text() is a promise, this prints` [object Promise]`

src/utils/mailup.ts
```diff
@@ -109,20 +106,29 @@ function callMailUpApi(
...
           )
         );
       }
-      return right<Error, ApiResponse>(response.body);
+      return ApiResponse.decode(response.json()).mapLeft(
```
gunzip, line 124: response.json is a promise

PR #71, opened 2020-06-08
[#172973044] Adds a nodemailer transport that sends emails through multiple transports

cloudify:


src/utils/__tests__/nodemailer.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    );
+
+    const result = await Promise.all(
+      new Array(100).fill(0).map(_ => multi.sendMail(anEmailMessage))
+    );
+
+    expect(result).toHaveLength(100);
+    expect(t1.send).toHaveBeenCalled();
+    expect(t2.send).toHaveBeenCalled();
+    expect(t3.send).toHaveBeenCalled();
```
gunzip, line 83: is this deterministic ? (since it's random, t1 may be not called)

cloudify: not deterministic but very very very unlikely :)

cloudify: I just calculated the probability of that happening, it's ~2.46 ∙ 10⁻¹⁸

PR #72, opened 2020-06-09
[#172973044] Adds types for configuring nodemailer MultiTransport

cloudify:


src/utils/multi_transport_connection.ts
```diff
@@ -0,0 +1,75 @@
...
+export type MailMultiTransportConnections = t.TypeOf<
+  typeof MailMultiTransportConnections
+>;
+
+/**
+ * Parses a multi email provider connection string.
+ *
+ * Multiple providers can be specified as following:
+ *
+ * transport:username:password[;transport:username:pasword][;transport:username:pasword]...
```
gunzip, line 32:
```suggestion
 * transport:username:password[;transport:username:password][;transport:username:password]...
```

gunzip: what's 'transport' in this context ? the name of nodemailer transport ?

src/utils/multi_transport_connection.ts
```diff
@@ -0,0 +1,75 @@
...
+          username
+        };
+      }
+      return undefined;
+    })
+    .filter(_ => _ !== undefined) as MailMultiTransportConnections;
+}
+
+/**
+ * Decodes an arrat of nodemailer transport connections from a multi
```
gunzip, line 59:
```suggestion
 * Decodes an array of nodemailer transport connections from a multi
```

src/utils/multi_transport_connection.ts
```diff
@@ -0,0 +1,75 @@
...
+        username !== undefined && // allow empty username
+        password !== undefined // allow empty password
+      ) {
+        return {
+          password,
+          transport,
+          username
+        };
+      }
+      return undefined;
```
gunzip, line 53: nit: if you want to avoid the cast here you can return an option instead of undefined and call catOptions on the resulting array: https://github.com/gcanti/fp-ts/blob/1.17.0/src/Array.ts#L1311

src/utils/multi_transport_connection.ts
cloudify: yes

PR #73, opened 2020-06-15
[#173218638] Handle promise rejection in mailup transport

cloudify:


src/utils/mailup.ts
```diff
@@ -293,21 +293,20 @@ export function MailUpTransport(
...
-      if (isRight(errorOrResponse)) {
-        // tslint:disable-next-line:no-null-keyword
-        return callback(null, {
-          ...errorOrResponse.value,
-          messageId: mail.data.messageId
-        });
-      } else {
-        return callback(errorOrResponse.value, undefined);
-      }
+      sendTransactionalMail(options.creds, email, fetchAgent)
```
gunzip, line 296: just note that `sendTransactionalMail` cannot throw any exception

cloudify: you never know :)

gunzip, line 296: do we need a `return` here ?

cloudify: nope, [the caller](https://github.com/nodemailer/nodemailer/blob/7750e048ff79d099500061497fb45b2861a7652c/lib/mailer/index.js#L214) doesn't expect a return value

PR #74, opened 2020-06-16
[#173351990] fromQueryEither

balanza:
## Motivation
Our db interfaces return results in the unpleasant form `Promise<Either<QueryError, T>>` while a form `TaskEither<QueryError, T>` would be preferrable. `fromQueryEither` is a helper method that does such type conversion.
This helper [has been used extensively in our codebase](https://github.com/search?q=fromQueryEither+user%3Apagopa&type=Code&ref=advsearch&l=&l=) and it needs to be defined as a common util to avoid duplication.

## Implementation
- Cosmos' query errors are returned as left 
- Runtime errors (that is: when the promise rejects for unexpected problems) are mapped in the form `{code: "error", body: error.message}` to mock the interface of Cosmos' QueryError type
- a convenient `DocumentDbUtils.QueryError` type is defined to map such union 

## Usage
see [test snippets](https://github.com/pagopa/io-functions-commons/blob/1beb3a112c55111f7633505aeec41326df863fa2/src/utils/__tests__/documentdb.test.ts#L676).

### DIFF
--- a/src/utils/__tests__/documentdb.test.ts
+++ b/src/utils/__tests__/documentdb.test.ts
@@ -5,9 +5,10 @@ import { NonEmptyString } from "italia-ts-commons/lib/strings";
 
 import * as DocumentDb from "documentdb";
 
-import { isLeft, isRight, left, right } from "fp-ts/lib/Either";
+import { Either, isLeft, isRight, left, right } from "fp-ts/lib/Either";
 import { none, some } from "fp-ts/lib/Option";
 
+import { TaskEither } from "fp-ts/lib/TaskEither";
 import * as DocumentDbUtils from "../documentdb";
 
 describe("getDatabaseUri", () => {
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/documentdb.test.ts
```diff
@@ -671,3 +672,71 @@ describe("queryAttachments", () => {
...
+  });
+
+  it("should return the correct query error on rejection", async () => {
+    interface IExpectedResultType {
+      foo: string;
+    }
+
+    const errorMessage = "generic error";
+    const queryFn: () => Promise<
+      Either<DocumentDb.QueryError | Error, IExpectedResultType>
```
gunzip, line 727: It looks like this promise here (the queryfn) cannot never return `left(Error)`

```suggestion
      Either<DocumentDb.QueryError, IExpectedResultType>
```

src/utils/documentdb.ts
```diff
@@ -790,4 +791,31 @@ export function queryAttachments<T>(
...
+
+/**
+ * Converts a Promise<Either> into a TaskEither
+ * This is needed because our models return unconvenient type. Both left and rejection cases are handled as a TaskEither left
+ * @param lazyPromise a lazy promise to convert
+ *
+ * @returns either the query result or a query failure
+ */
+export const fromQueryEither = <R>(
+  lazyPromise: () => Promise<Either<DocumentDb.QueryError | Error, R>>
```
gunzip, line 809: Why the signature of this promise is 

`Promise<Either<DocumentDb.QueryError | Error, R>>` and not just

`Promise<Either<DocumentDb.QueryError, R>>` ?

balanza: For a more relaxed API, but I think we can be more specific

src/utils/documentdb.ts
```diff
@@ -790,4 +791,31 @@ export function queryAttachments<T>(
...
 
+/**
+ * A convenient type that extends Cosmos' DocumentDb.QueryError with a generic status that maps a runtime error (instance of Error)
+ */
+export type QueryError =
+  | DocumentDb.QueryError
+  | { code: "error"; body: string };
+
+/**
+ * Converts a Promise<Either> into a TaskEither
```
cloudify, line 802:
```suggestion
 * Converts a Promise<Either<L, R>> into a TaskEither<L, R>
```

PR #75, opened 2020-06-17
[#173351990]  better query error conversion

balanza:
### Motivation
After using the features from #74 I realised that, in case of failure, `toError` would force a conversion from `DocumentDb.QueryError` to `Error`, resulting in a non-readable message.

In order to help troubleshooting, I introduced a `toQueryError` function to be used instead

### Usage
Actually, it's used internally by `fromQueryEither` without changing its signature. It's just a more readable error.


gunzip:
needs rebasing

src/utils/__tests__/documentdb.test.ts
```diff
@@ -739,4 +739,58 @@ describe("fromQueryEither", () => {
...
+    }
+
+    const cosmosdbError: DocumentDb.QueryError = {
+      body: "Bad Request",
+      code: 400
+    };
+    const queryFn: () => Promise<
+      Either<DocumentDb.QueryError, IExpectedResultType>
+    > = async () => {
+      throw cosmosdbError;
```
gunzip, line 755: it's not clear, AFAIK the cosmosdb error is never throw but returned in the left side of the either

balanza:
No clear benefit, closing

PR #77, opened 2020-06-30
Migrating to cosmos sdk v3

cloudify:


src/utils/documentdb_model.ts
```diff
@@ -47,25 +49,22 @@ export abstract class DocumentDbModel<
    *     resolves to either an error or the retrieved version of the persisted
    *     document.
    */
-  public async create(
-    document: TN,
-    partitionKey: string
-  ): Promise<Either<DocumentDb.QueryError, TR>> {
+  public create(document: TN): TaskEither<ErrorResponse, TR> {
     // we don't want to store the "kind" property, let's remove it
+    // FIXME: is this filter still needed?
```
cloudify, line 54: @gunzip @francescopersico do we still need to remove the `kind` attribute before storing objects on cosmos?

src/utils/documentdb_model.ts
```diff
@@ -47,51 +83,53 @@ export abstract class DocumentDbModel<
...
    * with the provided id.
    *
    */
-  public async createOrUpdate(
+  public createOrUpdate(
     document: TN,
-    partitionKey: string
-  ): Promise<Either<DocumentDb.QueryError, TR>> {
+    options?: RequestOptions
+  ): TaskEither<ErrorResponse, TR> {
```
francescopersico, line 119: Why ErrorResponse here instead of QErrors?

src/utils/documentdb_model.ts
```diff
@@ -47,51 +83,53 @@ export abstract class DocumentDbModel<
...
-      this.collectionUri,
-      kindlessDocument,
-      partitionKey
-    );
-
-    // if the result is successful we map it to a TR type
-    return maybeCreatedOrUpdatedDocument.map(this.toRetrieved);
+    return tryCatch(
+      () => this.container.items.upsert(kindlessDocument, options),
+      toError
```
francescopersico, line 128:
```
      e => ({
        error: toError(e),
        kind: "RESPONSE"
      })
```
Like above?

src/utils/documentdb_model.ts
```diff
@@ -100,32 +138,15 @@ export abstract class DocumentDbModel<
...
-      // if the error is 404 (Not Found), we return an empty value
-      return right<DocumentDb.QueryError, Option<TR>>(none);
-    }
-
-    // for any other error (errorOrDocument is a left), we return it as is
-    // or in case of success, we map the result to a retrieved interface
-    return errorOrDocument.map(r => some(this.toRetrieved(r)));
+    partitionKey: string,
+    options?: RequestOptions
+  ): TaskEither<ErrorResponse, Option<TR>> {
```
francescopersico, line 145: Same here

cloudify:
This was a previous iteration, last one is only for the create method, once we agree on that I'll update the rest

src/utils/documentdb_model.ts
```diff
@@ -1,9 +1,51 @@
...
 
-import * as DocumentDbUtils from "./documentdb";
+type TN<T> = t.Type<T, ItemDefinition>;
 
-import { Either, isLeft, left, right } from "fp-ts/lib/Either";
-import { fromNullable, none, Option, some } from "fp-ts/lib/Option";
+type TR<T> = t.IntersectionC<
+  // tslint:disable-next-line: readonly-array
+  [t.Type<T, ItemDefinition>, typeof ResourceT]
+>;
```
balanza, line 48: Should we prefer meaningful names like `RetrievedDocument` and `NewDocument`?

cloudify:
Yes definitely, this is still very very draft

src/utils/documentdb_model.ts
cloudify: let me know what do you think about how's `create` structured

src/utils/documentdb_model.ts
```diff
@@ -47,51 +83,53 @@ export abstract class DocumentDbModel<
...
+    const item = this.type.encode(document);
+    return tryCatch<QErrors, PromiseType<ReturnType<typeof createItem>>>(
+      () => createItem(item, options),
+      e => ({
+        error: toError(e),
+        kind: "RESPONSE"
+      })
+    )
+      .map(_ => _.resource)
+      .filterOrElse(isDefined, { kind: "EMPTY" }) // TODO: not sure whether an undef resource should be an error
```
AleDore, line 100: Maybe for a new document, we should return an error, because it' s similar to insert a tuple with all columns set to null, in the relational world isn't it?

cloudify: that's what I don't know yet, I suspect `undefined` is part of the response type used for all methods but not actually returned by `create`, have to check the sdk code to confirm

gunzip:
lgtm so far

src/utils/async.ts
```diff
@@ -0,0 +1,15 @@
+export function mapAsyncIterable<T, V>(
+  source: AsyncIterable<T>,
+  f: (t: T) => V
+): AsyncIterable<V> {
+  const iter = source[Symbol.asyncIterator]();
+  return {
+    [Symbol.asyncIterator]: () => ({
+      next: () =>
+        iter.next().then(({ done, value }) => ({
```
gunzip, line 9: what if next() throws?

cloudify: I guess the iterator will stop iterating? it looks like that's the expected behavior - what do you suggest to do?

src/utils/cosmosdb_model.ts
```diff
@@ -0,0 +1,188 @@
+// tslint:disable: member-ordering
+
+import { right } from "fp-ts/lib/Either";
```
cloudify, line 3: @pagopa/contributors migration of the base cosmos model is pretty much done, please check `cosmosdb_model` and `cosmosdb_model_versioned` - I'm going to finish writing tests and migrating the existing models

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    return super.create(newDocument, options);
+  };
+
+  /**
+   * Creates a new version from a full item definition
+   */
+  public upsert = (o: T): TaskEither<CosmosErrors, TR> => {
+    const modelId = this.getModelId(o);
+    // FIXME: we shouldn't need to inc the version here, `o` should have already the nextVersion
```
gunzip, line 112: so this is left only for backward compatibility isn't it?

src/utils/cosmosdb_model.ts
```diff
@@ -0,0 +1,188 @@
...
+   */
+  public find(
+    documentId: string,
+    partitionKey: string,
+    options?: RequestOptions
+  ): TaskEither<CosmosErrors, Option<TR>> {
+    const readItem = this.container.item(documentId, partitionKey).read;
+    return tryCatch<CosmosErrors, PromiseType<ReturnType<typeof readItem>>>(
+      () => readItem(options),
+      _ => CosmosErrorResponse(_ as ErrorResponse)
```
gunzip, line 162: ditto

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -0,0 +1,192 @@
...
+   * Returns the value of the model ID for the provided item
+   */
+  protected getModelId = (o: T) =>
+    ModelId.decode(String(o[this.modelIdKey])).value as ModelId;
+
+  /**
+   * Returns the most recent version of the model from the database or 0 if
+   * no previous version is found.
+   */
+  private getLastVersion = (modelId: ModelId) =>
```
gunzip, line 186: the name is a little bit misleading since this not get the latest version but the next one (which is called "current" above)

this is hard to grasp: getLastVersion = findLastVersionByModelId + 1

cloudify: ops you are right, it should be called `getNextVersion`

src/utils/async.ts
gunzip: the important thing here is that the exception is catchable otherwise it may stop the whole process

cloudify: I'm not sure it should be a concern of this mapping function, `iter.next()` could throw even when not mapping over it (i.e. the underlying iterator could throw) - how would you handle it?

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -0,0 +1,190 @@
...
+ * @param version The version of the model
+ */
+export function generateVersionedModelId(
+  modelId: ModelId,
+  version: NonNegativeInteger
+): string {
+  const paddingLength = 16; // length of Number.MAX_SAFE_INTEGER == 9007199254740991
+  const paddedVersion = ("0".repeat(paddingLength) + String(version)).slice(
+    -paddingLength
+  );
```
balanza, line 58: We can use `padStart` ([safe with node >= 8.2](https://node.green/#ES2017-features-String-padding-String-prototype-padStart))

```suggestion
  const paddedVersion = String(version).padStart(paddingLength, "0");
```

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  public create = (
+    o: TN,
+    options?: RequestOptions
+  ): TaskEither<CosmosErrors, TR> => {
+    // if we get an explicit version number from the new document we use that,
+    // or else we set version to 0
+    const currentVersion: NonNegativeInteger | undefined = o.version;
+    const version =
+      currentVersion === undefined ? (0 as NonNegativeInteger) : currentVersion;
```
balanza, line 94: Ensure is NonNegativeInteger

```suggestion
    const version = NonNegativeInteger.decode(o.version).getOrElse(
      0 as NonNegativeInteger
    );
```

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -0,0 +1,190 @@
...
+    // this makes it possible to detect conflicting updates (concurrent creation of
+    // profiles with the same profile ID and version)
+    const modelId = this.getModelId(o);
+    const versionedModelId = generateVersionedModelId(modelId, version);
+
+    const newDocument = {
+      ...o,
+      id: versionedModelId,
+      version
+    } as TN & VersionedModel & BaseModel;
```
balanza, line 106: Should we prefer `encode` to cast? The following code has errors, but it might be a lead:

```typescript
    const newDocument = t
      .intersection([this.newVersionedItemT, VersionedModel, BaseModel])
      .encode({
        ...o,
        id: versionedModelId,
        version
      });
```

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -0,0 +1,190 @@
...
+   */
+  public upsert = (o: TN): TaskEither<CosmosErrors, TR> => {
+    // if we get an explicit version number from the new document we use that,
+    // or else we get the last version by querying the database
+    const currentVersion: NonNegativeInteger | undefined = o.version;
+    const modelId = this.getModelId(o);
+    return (currentVersion === undefined
+      ? this.getNextVersion(modelId)
+      : fromEither<CosmosErrors, NonNegativeInteger>(right(currentVersion))
+    ).chain(nextVersion =>
```
balanza, line 122: Ensure is NonNegativeInteger

```suggestion
    return NonNegativeInteger.decode(currentVersion)
      .fold(
        _ => this.getNextVersion(modelId),
        _ => taskEither.of(_)
      )
      .chain(nextVersion =>
```

balanza, line 122: I think it's a bit misleading: `nextVersion` sounds like it's `currentVersion + 1`, but it's potentially the very same value. 
Should we change `currentVersion` with `inputVersion`?

balanza:
@cloudify I proposed some test refactor here: https://github.com/pagopa/io-functions-commons/pull/79

The point is it may be quicker to write more tests

src/utils/async.ts
```diff
@@ -0,0 +1,50 @@
...
+    [Symbol.asyncIterator]: () => iterMapped
+  };
+}
+
+export async function asyncIteratorToArray<T>(
+  iter: AsyncIterator<T>
+): Promise<ReadonlyArray<T>> {
+  const acc = Array<T>();
+
+  while (true) {
```
AleDore, line 36: Is it possible to use hasMoreResults or similar?

src/models/message.ts
```diff
@@ -257,23 +229,60 @@ export class MessageModel extends DocumentDbModel<
...
+    return fromEitherT(
+      tryCatch2v(
+        () => iterator,
+        _ => CosmosErrorResponse(_ as ErrorResponse)
+      )
     );
   }
 
+  /**
+   * @deprecated use getQueryIterator + asyncIterableToArray
```
gunzip, line 255: so why leave it?

src/models/notification.ts
```diff
@@ -194,20 +179,45 @@ export class NotificationModel extends DocumentDbModel<
...
+          name: "@messageId",
+          value: messageId
+        }
+      ],
+      query: `SELECT * FROM n WHERE (n.${NOTIFICATION_MODEL_PK_FIELD} = @messageId)`
+    });
+  }
+
+  /**
+   * @deprecated use getQueryIterator + asyncIterableToArray
```
gunzip, line 195: ditto

src/models/notification.ts
```diff
@@ -194,20 +180,42 @@ export class NotificationModel extends DocumentDbModel<
...
+      .chain(_ =>
+        _.isSome()
+          ? fromEitherT(
+              RetrievedNotification.decode(_.value[0])
+                .map(some)
+                .mapLeft(CosmosDecodingError)
+            )
+          : fromEitherT(rightE(none))
+      );
   }
```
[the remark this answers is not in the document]
cloudify: yes definitely - @AleDore can you refactor?

src/models/message.ts
cloudify: I didn't wanted to change too many things in the functions code - indeed I would rather remove the code, is that ok with you?

gunzip: yes

src/models/notification.ts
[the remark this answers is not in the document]
AleDore: Of course

src/utils/__tests__/cosmosdb_model.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    containerMock.items.create.mockRejectedValueOnce({ statusCode: 500 });
+    const model = new MyModel(container);
+
+    const result = await model.create(aDocument).run();
+    expect(isLeft(result));
+    if (isLeft(result)) {
+      expect(result.value.kind).toBe("COSMOS_ERROR_RESPONSE");
+      if (
+        result.value.kind === "COSMOS_ERROR_RESPONSE" &&
+        fromNullable(result.value.error.code).isSome()
```
cloudify, line 94: @AleDore this branch gets never executed as `code` is always `undefined`, `create` should return an `ErrorResponse` that has `code: 500`

src/utils/__tests__/cosmosdb_model.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    if (isRight(result)) {
+      expect(result.value).toEqual({
+        ...aDocument,
+        ...someMetadata
+      });
+    }
+  });
+
+  it("should fail on query error", async () => {
+    containerMock.items.create.mockRejectedValueOnce({ statusCode: 500 });
```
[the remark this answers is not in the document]
AleDore: Yes I'll correct this

src/utils/__tests__/cosmosdb_model.test.ts
AleDore: It depends from previous wrong mock, but I've noticed that `error.code` is always undefined, even if you mock it with `{code: 500}`

src/utils/async.ts
AleDore: Nevermind, I was looking for another way to iterate over next elements of the iterator, but it seems to be the only way.

src/utils/cosmosdb_model.ts
```diff
@@ -0,0 +1,267 @@
...
+    options?: FeedOptions
+  ): TaskEither<CosmosErrors, Option<TR>> {
+    const queryIterator = this.container.items.query<TR>(query, options)
+      .fetchAll;
+    return tryCatch<
+      CosmosErrors,
+      PromiseType<ReturnType<typeof queryIterator>>
+    >(
+      () => queryIterator(),
+      _ => CosmosErrorResponse(_ as ErrorResponse)
```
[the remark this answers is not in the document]
AleDore: Ok I'll refactor

src/models/message.ts
```diff
@@ -257,23 +229,60 @@ export class MessageModel extends DocumentDbModel<
...
+        _.isSome()
+          ? fromEitherT(
+              array.sequence(either)(
+                _.value.map(RetrievedMessageWithoutContent.decode)
+              )
+            )
+              .map(some)
+              .mapLeft(CosmosDecodingError)
+          : fromEitherT(right(none))
+      );
```
gunzip, line 283: this should go into the parent model since it's a generic method. can we rewrite findMessages using this one?

src/models/message_status.ts
```diff
@@ -109,61 +57,39 @@ export const getMessageStatusUpdater = (
...
     messageId: NonEmptyString
-  ): Promise<Either<DocumentDb.QueryError, Option<RetrievedMessageStatus>>> {
-    return super.findLastVersionByModelId(
-      MESSAGE_STATUS_MODEL_ID_FIELD,
-      messageId,
-      MESSAGE_STATUS_MODEL_PK_FIELD,
-      messageId
-    );
+  ): TaskEither<CosmosErrors, Option<RetrievedMessageStatus>> {
+    return super.findLastVersionByModelId(messageId, messageId);
```
gunzip, line 93: it looks like we can safely remove this method which is just an alias

src/models/user_data_processing.ts
```diff
@@ -158,61 +98,46 @@ export class UserDataProcessingModel extends DocumentDbModelVersioned<
...
     );
   }
 
   /**
    * Searches for one user data processing request to the provided id
    *
    * @param id
+   * @deprecated Use findLastVersionByModelId(userDataProcessingId, fiscalCode)
    */
   public findOneUserDataProcessingById(
```
gunzip, line 116: it looks like we can safely remove this method which is just an alias

src/models/message.ts
```diff
@@ -257,23 +229,60 @@ export class MessageModel extends DocumentDbModel<
...
+          value: fiscalCode
+        }
+      ],
+      query: `SELECT * FROM m WHERE m.${MESSAGE_MODEL_PK_FIELD} = @fiscalCode`
+    })[Symbol.asyncIterator]();
+
+    return fromEitherT(
+      tryCatch2v(
+        () => iterator,
+        _ => CosmosErrorResponse(_ as ErrorResponse)
```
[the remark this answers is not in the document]
AleDore: yes

src/models/message.ts
```diff
@@ -257,23 +229,60 @@ export class MessageModel extends DocumentDbModel<
...
+    CosmosErrors,
+    Option<ReadonlyArray<RetrievedMessageWithoutContent>>
+  > {
+    const fetchAll = this.container.items.query<RetrievedMessageWithoutContent>(
+      query,
+      options
+    ).fetchAll;
+    return tryCatchT<CosmosErrors, PromiseType<ReturnType<typeof fetchAll>>>(
+      () => fetchAll(),
+      _ => CosmosErrorResponse(_ as ErrorResponse)
```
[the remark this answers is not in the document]
AleDore: yes

src/models/message.ts
AleDore: It' s deprecated, cause we have to use `findMessages` using `AsyncIterator`. Maybe It can be an other option to implement this method in a generic way for other purposes @cloudify

src/models/message.ts
```diff
@@ -239,11 +206,11 @@ export class MessageModel extends DocumentDbModel<
...
-  public async findMessageForRecipient(
+  public findMessageForRecipient(
     fiscalCode: FiscalCode,
     messageId: string
-  ): Promise<Either<DocumentDb.QueryError, Option<RetrievedMessage>>> {
-    const errorOrMaybeMessage = await this.find(messageId, fiscalCode);
+  ): TaskEither<CosmosErrors, Option<RetrievedMessage>> {
+    const errorOrMaybeMessage = super.find(messageId, fiscalCode);
 
     return errorOrMaybeMessage.map(maybeMessage =>
```
gunzip, line 215: why do we need this filter here ? isn't the fiscalCode already filtered by the query?

src/utils/__tests__/cosmosdb_model.test.ts
cloudify: Yes the mock doesn't work, the method toCosmosErrors doesn't recognize it as an ErrorRrsponse, the mock should return an ErrorResponse

src/models/message.ts
cloudify: I asked myself the same question, I believe it's there as a "double check to be sure", not a big deal though

PR #81, opened 2020-07-27
[#171554236] Document IDs as NonEmptyString

balanza:
Enforce document IDs as NonEmptyString. This is to avoid manual casting when using objects coming from db (example: fetch a document, edit a field, save the document).

The assumption is safe as `id` field is never empty on Cosmos document. 

What we are losing with these changes is the direct relation between the decoder `ResourceT` and the type `@azure/cosmos/Resource`. This has been mitigated by the introduction of type `ResourceT` and [this cast](https://github.com/pagopa/io-functions-commons/blob/11ae8056e02750220c50eec905f04de1f42a0ae9/src/utils/cosmosdb_model.ts#L45).

`ResourceT` is actually an extension of `@azure/cosmos/Resource`, and hence we should prefer the former to the latter. I see we don't use such kind explicitly in our codebase, though.

### DIFF
--- a/src/utils/cosmosdb_model.ts
+++ b/src/utils/cosmosdb_model.ts
@@ -16,21 +16,22 @@ import {
   ItemDefinition,
   ItemResponse,
   RequestOptions,
-  Resource,
   SqlQuerySpec
 } from "@azure/cosmos";
 
+import { NonEmptyString } from "italia-ts-commons/lib/strings";
 import { mapAsyncIterable } from "./async";
 import { isDefined } from "./types";
 
 export const BaseModel = t.interface({
-  id: t.string // FIXME: should this be a NonEmptyString?
+  id: NonEmptyString
 });
 
 export type BaseModel = t.TypeOf<typeof BaseModel>;
 
 // An io-ts definition of Cosmos Resource runtime type
-// tslint:disable-next-line: no-useless-cast
+// IDs are enforced to be non-empty string, as we're sure they are alway valued when coming from db.
+export type ResourceT = t.TypeOf<typeof ResourceT>;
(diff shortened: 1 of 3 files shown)


src/utils/cosmosdb_model.ts
```diff
@@ -20,16 +20,19 @@ import {
...
 
 export const BaseModel = t.interface({
-  id: t.string // FIXME: should this be a NonEmptyString?
+  id: NonEmptyString
 });
 
 export type BaseModel = t.TypeOf<typeof BaseModel>;
 
 // An io-ts definition of Cosmos Resource runtime type
+// IDs are enforced to be non-empty string, as we're sure they are alway valued when coming from db.
```
cloudify, line 34:
```suggestion
// IDs are enforced to be non-empty string, as we're sure they are always valued when coming from db.
```

src/utils/cosmosdb_model.ts
```diff
@@ -39,7 +42,7 @@ export const ResourceT = t.intersection([
     _ts: t.number
   }),
   BaseModel
-]) as t.Type<Resource>;
+]) as t.Type<Resource & { id: NonEmptyString }>; // this cast is needed to keep ResourceT in sync with Resource (try to remove a field from the decoder definition and you'll face an error)
```
cloudify, line 45: perhaps

```suggestion
]) as Omit<t.Type<Resource, "id"> & { id: NonEmptyString }>; // this cast is needed to keep ResourceT in sync with Resource (try to remove a field from the decoder definition and you'll face an error)
```

balanza: Well, the `id` attribute would be override by the one on the right

PR #82, opened 2020-07-28
Add AsyncIterator / Iterable map, reduces, flatten

BurnedMarshal:
Allow mapping AsyncIterator with Either values
Allow flat AsyncIteretor with Readonly Array

### DIFF
--- a/src/utils/async.ts
+++ b/src/utils/async.ts
@@ -49,6 +49,13 @@ export async function asyncIterableToArray<T>(
   return asyncIteratorToArray(iter);
 }
 
+/**
+ * Create a new AsyncIterable providing only the values that satisfy the predicate function.
+ * The predicate function is also an optional Type Guard function if types T and K are different.
+ *
+ * @param iterable Original AsyncIterable
+ * @param predicate Predicate function
+ */
 export const filterAsyncIterable = <T, K = T>(
   iterable: AsyncIterable<T | K>,
   predicate: (value: T | K) => value is K
@@ -63,6 +70,13 @@ export const filterAsyncIterable = <T, K = T>(
   }
 });
 
+/**
+ * Create a new AsyncIterator providing only the values that satisfy the predicate function.
+ * The predicate function is also an optional Type Guard function if types T and K are different.
+ *
+ * @param iter Original AsyncIterator
+ * @param predicate Predicate function
+ */
 export function filterAsyncIterator<T, K = T>(
   iter: AsyncIterator<T | K>,
   predicate: (value: T | K) => value is K
@@ -73,6 +87,11 @@ export function filterAsyncIterator<T, K = T>(
   return filterAsyncIterable(iterable, predicate)[Symbol.asyncIterator]();
 }
 
+/**
+ * Create a new AsyncIterable which provide one by one the values ​​contained into the input AsyncIterable
+ *
+ * @param iterable Original AsyncIterable
+ */
 export const flattenAsyncIterable = <T>(
   iterable: AsyncIterable<ReadonlyArray<T>>
 ): AsyncIterable<T> => ({
@@ -86,6 +105,11 @@ export const flattenAsyncIterable = <T>(
   }
 });
 
+/**
+ * Create a new AsyncIterator which provide one by one the values ​​contained into the input AsyncIterator
+ *
+ * @param iter Original AsyncIterator
+ */
 export function flattenAsyncIterator<T>(
   iter: AsyncIterator<ReadonlyArray<T>>
 ): AsyncIterator<T> {
--- a/src/utils/async.ts
+++ b/src/utils/async.ts
@@ -79,14 +79,18 @@ export function flattenAsyncIterator<T>(
   let flattenArray: ReadonlyArray<T> = [];
   return {
     next: async () => {
-      if (flattenArray.length === index) {
-        const { done, value } = await iter.next();
-        if (done) {
-          return { done, value: undefined };
+      while (true) {
+        if (flattenArray.length === index) {
+          const { done, value } = await iter.next();
+          if (done) {
+            return { done, value: undefined };
+          }
+          flattenArray = value;
+          index = 0;
+          continue;
         }
-        flattenArray = [...flattenArray, ...value];
+        return { done: false, value: flattenArray[index++] };
       }
-      return { done: false, value: flattenArray[index++] };
     }
   };
 }
--- a/src/utils/async.ts
+++ b/src/utils/async.ts
@@ -53,6 +53,11 @@ export async function asyncIterableToArray<T>(
  * Create a new AsyncIterable providing only the values that satisfy the predicate function.
  * The predicate function is also an optional Type Guard function if types T and K are different.
  *
+ * Example:
+ * ```
+ * const i: AsyncIterable<Either<E, A>> = {} as AsyncIterable<Either<E, A>>;
+ * const f: AsyncIterable<Right<E, A>> = filterAsyncIterable<Either<E, A>, Right<E, A>>(i, isRight);
+ * ```
  * @param iterable Original AsyncIterable
  * @param predicate Predicate function
  */
@@ -74,6 +79,11 @@ export const filterAsyncIterable = <T, K = T>(
  * Create a new AsyncIterator providing only the values that satisfy the predicate function.
  * The predicate function is also an optional Type Guard function if types T and K are different.
  *
+ * Example:
+ * ```
+ * const i: AsyncIterator<Either<E, A>> = {} as AsyncIterator<Either<E, A>>;
+ * const newI: AsyncIterator<Right<E, A>> = filterAsyncIterator<Either<E, A>, Right<E, A>>(iterable, isRight);
+ * ```
  * @param iter Original AsyncIterator
  * @param predicate Predicate function
  */
--- a/src/utils/async.ts
+++ b/src/utils/async.ts
@@ -51,10 +51,10 @@ export async function asyncIterableToArray<T>(
   return asyncIteratorToArray(iter);
 }
 
-export function mapEitherAsyncIterator<E, T, V>(
-  iter: AsyncIterator<Either<E, T>, Either<E, T>>,
-  f: (t: T) => V
-): AsyncIterator<V> {
+export function filterAsyncIterator<T>(
+  iter: AsyncIterator<T>,
+  predicate: (t: T) => boolean
+): AsyncIterator<T> {
   return {
     next: async () => {
       while (true) {
@@ -62,8 +62,8 @@ export function mapEitherAsyncIterator<E, T, V>(
         if (done) {
           return { done, value: undefined };
         }
-        if (isRight(value)) {
-          return { done, value: f(value.value) };
+        if (predicate(value)) {
+          return { done, value };
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/utils/async.ts
```diff
@@ -48,3 +50,42 @@ export async function asyncIterableToArray<T>(
...
+  };
+}
+
+export async function flattenAsyncIterator<T>(
+  iter: AsyncIterator<ReadonlyArray<T>>
+): Promise<AsyncIterator<T>> {
+  const array = await asyncIteratorToArray(iter);
+  const flattenArray = array.reduce((acc, item) => {
+    return acc.concat(item);
+  }, []);
```
gunzip, line 79: doesn't this return the same as `array`? (it looks like an identity)

BurnedMarshal: With the `concat` method 2 ore more arrays are joined into flattenArray. It will be an identity if `push` was used instead of `concat`. Maybe is better `return [...acc, ...item]`?

BurnedMarshal:
@gunzip here an alternative sync version of `flattenAsyncIterator`
```typescript
export function flattenAsyncIterator<T>(
  iter: AsyncIterator<ReadonlyArray<T>>
): AsyncIterator<T> {
  // tslint:disable-next-line: no-let
  let index = 0;
  // tslint:disable-next-line: no-let
  let flattenArray: ReadonlyArray<T> = [];
  return {
    next: async () => {
      if (flattenArray.length === index) {
        const { done, value } = await iter.next();
        if (done) {
          return { done, value: undefined };
        }
        flattenArray = [...flattenArray, ...value];
      }
      return { done: false, value: flattenArray[index++] };
    }
  };
}
```

src/utils/async.ts
```diff
@@ -48,3 +50,43 @@ export async function asyncIterableToArray<T>(
   const iter = source[Symbol.asyncIterator]();
   return asyncIteratorToArray(iter);
 }
+
+export function mapEitherAsyncIterator<E, T, V>(
+  iter: AsyncIterator<Either<E, T>, Either<E, T>>,
```
cloudify, line 55: this looks too specific, I would implement a `filterAsyncIterator` instead:

```ts
filterAsyncIterator<T>(
  iter: AsyncIterator<T>,
  predicate: (t: T) => boolean
): AsyncIterator<T>;
```

then `mapEitherAsyncIterator` becomes:

```ts
filterAsyncIterator(mapAsyncIterator(iter, _ => _.isRight() ? _.value : undefined), isDefined);
```

where `isDefined` is a guard

src/utils/async.ts
```diff
@@ -48,3 +50,43 @@ export async function asyncIterableToArray<T>(
...
+  // tslint:disable-next-line: no-let
+  let index = 0;
+  // tslint:disable-next-line: no-let
+  let flattenArray: ReadonlyArray<T> = [];
+  return {
+    next: async () => {
+      if (flattenArray.length === index) {
+        const { done, value } = await iter.next();
+        if (done) {
+          return { done, value: undefined };
```
cloudify, line 85: the caller should probably be aware that the last returned value gets ignored

src/utils/async.ts
```diff
@@ -48,3 +50,43 @@ export async function asyncIterableToArray<T>(
...
+        }
+        flattenArray = [...flattenArray, ...value];
+      }
+      return { done: false, value: flattenArray[index++] };
```
cloudify, line 89: I believe you should check again whether `index < flattenArray.length` here as `value` returned by `next()` could be an empty array

src/utils/async.ts
```diff
@@ -48,3 +50,43 @@ export async function asyncIterableToArray<T>(
...
+  // tslint:disable-next-line: no-let
+  let flattenArray: ReadonlyArray<T> = [];
+  return {
+    next: async () => {
+      if (flattenArray.length === index) {
+        const { done, value } = await iter.next();
+        if (done) {
+          return { done, value: undefined };
+        }
+        flattenArray = [...flattenArray, ...value];
```
cloudify, line 87: this approach makes `flattenArray` keep growing, instead you could only keep the last array:

```ts
flattenArray = value;
index = 0;
```

src/utils/async.ts
BurnedMarshal: Fixed in 8f4c631

src/utils/async.ts
BurnedMarshal: Fixed in 57b70cb

src/utils/async.ts
BurnedMarshal: Fixed in 57b70cb

src/utils/async.ts
BurnedMarshal: How do you suggest to warning the caller? Could be enough with a comment description over the function?

src/utils/async.ts
```diff
@@ -48,3 +48,49 @@ export async function asyncIterableToArray<T>(
   const iter = source[Symbol.asyncIterator]();
   return asyncIteratorToArray(iter);
 }
+
+export const filterAsyncIterable = <T, K = T>(
```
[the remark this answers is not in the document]
gunzip: ping :-)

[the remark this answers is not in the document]
BurnedMarshal: I'll add some comments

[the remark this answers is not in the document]
gunzip: I'd  add what T & K are supposed to be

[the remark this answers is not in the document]
BurnedMarshal: With an example?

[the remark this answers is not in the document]
gunzip: why not:)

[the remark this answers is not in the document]
BurnedMarshal: Fixed in 51a8c26 and 46aaaa0

src/utils/async.ts
```diff
@@ -48,3 +48,109 @@ export async function asyncIterableToArray<T>(
...
+ * ```
+ * const i: AsyncIterable<Either<E, A>> = {} as AsyncIterable<Either<E, A>>;
+ * const f: AsyncIterable<Right<E, A>> = filterAsyncIterable<Either<E, A>, Right<E, A>>(i, isRight);
+ * ```
+ * @param iterable Original AsyncIterable
+ * @param predicate Predicate function
+ */
+export const filterAsyncIterable = <T, K = T>(
+  iterable: AsyncIterable<T | K>,
+  predicate: (value: T | K) => value is K
```
cloudify, line 66: How does this work when value is not a union of more than one type? Say we have an AsyncIterable<string> and the predicate returns true for non empty strings, what are T and K?

src/utils/async.ts
```diff
@@ -48,3 +48,109 @@ export async function asyncIterableToArray<T>(
...
+ * const newI: AsyncIterator<Right<E, A>> = filterAsyncIterator<Either<E, A>, Right<E, A>>(iterable, isRight);
+ * ```
+ * @param iter Original AsyncIterator
+ * @param predicate Predicate function
+ */
+export function filterAsyncIterator<T, K = T>(
+  iter: AsyncIterator<T | K>,
+  predicate: (value: T | K) => value is K
+): AsyncIterator<K> {
+  const iterable = {
```
cloudify, line 94: it would be better to define the Iterable in terms of the Iterator, not the other way around

src/utils/async.ts
```diff
@@ -48,3 +48,109 @@ export async function asyncIterableToArray<T>(
...
+    }
+  }
+});
+
+/**
+ * Create a new AsyncIterator which provide one by one the values ​​contained into the input AsyncIterator
+ *
+ * @param iter Original AsyncIterator
+ */
+export function flattenAsyncIterator<T>(
```
cloudify, line 123: Same here, it would make more sense to define flattenAsyncIterable in terms of flattenAsyncIterator

src/utils/async.ts
```diff
@@ -48,3 +48,109 @@ export async function asyncIterableToArray<T>(
...
+  iter: AsyncIterator<ReadonlyArray<T>>
+): AsyncIterator<T> {
+  const iterable = {
+    [Symbol.asyncIterator]: () => iter
+  };
+  return flattenAsyncIterable(iterable)[Symbol.asyncIterator]();
+}
+
+export function reduceAsyncIterable<A, B>(
+  iterable: AsyncIterable<ReadonlyArray<A>>,
```
cloudify, line 133: This should take an AsyncIterable<A> (we can flatten AsyncIterable<ReadonlyArray<A>> to get an AsyncIterable<A>)

src/utils/async.ts
BurnedMarshal: It works with any input type, even if not an union. For example, this code transcompile successfully
```typescript
const i: AsyncIterator<unknown> = {} as AsyncIterator<unknown>;
  const newI: AsyncIterator<NonEmptyString> = filterAsyncIterator<unknown, NonEmptyString>(i, (_): _ is NonEmptyString => {
    if (isRight(NonEmptyString.decode(_))) {
      return true;
    }
    return false;
});
```

src/utils/async.ts
BurnedMarshal: If we flatten `AsyncIterable<ReadonlyArray>` the output will be different. This function reduces the single `ReadonlyArray` to one value of the AsyncIterator.
I think that this test explain what I mean [link](https://github.com/pagopa/io-functions-commons/pull/82/files#diff-bc0a6d43f117d3c7951c79b835270311R128-R152)

cloudify: ah I see, the name is misleading then, and you don't need a specific method to do that, you can map the iterator with a reduce:

```ts
mapAsyncIterator(iter, _ => _.reduce(f, zero))
```

src/utils/async.ts
balanza: glass half-full: this is unpractical, but it would force a more fine-grained modeling 
```ts
const RetrievedPendingMessage = t.intersection([
  RetrievedMessage,
  t.interface({ isPending: t.literal(true) })
]);

messageModel
    .findMessages(fiscalCode)
    .map(flattenAsyncIterator)
    .map(_ => filterAsyncIterator(_, RetrievedPendingMessage.is))
```

src/utils/async.ts
BurnedMarshal: Nice :), so I'll remove the code.

src/utils/async.ts
balanza: I realised the above code to be wrong. In 
```ts
.map(_ => filterAsyncIterator(_, RetrievedPendingMessage.is))
```
`_` is of type `AsyncIterator<Either<t.Errors, RetrievedMessage>>`, thus the predicate will take an Either.

Should `filterAsyncIterator` care about the value to be `left` or `right`?

BurnedMarshal: @balanza yes, you have an Either in input. the right code will be:
```typescript
messageModel
  .findMessages(fiscalCode)
  .map(flattenAsyncIterator)
  .map(_ =>
    filterAsyncIterator<
      Either<Errors, RetrievedMessage>,
      Right<Errors, RetrievedPendingMessage>
    >(
      _,
      (val): val is Right<Errors, RetrievedPendingMessage> =>
        isRight(val) && RetrievedPendingMessage.is(val)
    )
  );
```

BurnedMarshal: Maybe this signature is more compact and clear:
```typescript
export function filterAsyncIterator<T>(
  iter: AsyncIterator<unknown>,
  predicate: (value: unknown) => value is T
): AsyncIterator<T> {
```

cloudify: we probably need two signatures to support both non type-guards and type guards predicates, as in the standard library: https://github.com/microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L1176

PR #83, opened 2020-07-30
Include internal fields into CosmosDB RetrievedDocument model

balanza:
This PR is to discuss the type modelling on `CosmosdbModel` and `CosmosdbModelVersioned` classes. 

The point is to have retrieved models to extend `CosmosResource` (formerly: `ResourceT`) instead of just `BaseModel`. This will require retrieved models' instances to explicitly handle the following fields:
```
_etag: t.string,
_rid: t.string,
_self: t.string,
_ts: t.number
```


src/utils/conversions.ts
```diff
@@ -1,50 +0,0 @@
-import { NonEmptyString } from "italia-ts-commons/lib/strings";
```
gunzip, line 1: why delete this?

src/utils/messages.ts
```diff
@@ -37,10 +37,10 @@ export async function getChannelStatus(
   const errorOrMaybeStatus = await notificationStatusModel
     .findOneNotificationStatusByNotificationChannel(notificationId, channel)
     .run();
-  return fromEither(errorOrMaybeStatus)
+  const c = fromEither(errorOrMaybeStatus)
```
gunzip, line 40: why this change?

src/utils/cosmosdb_model.ts
```diff
@@ -32,17 +32,17 @@ export type BaseModel = t.TypeOf<typeof BaseModel>;
 
 // An io-ts definition of Cosmos Resource runtime type
 // IDs are enforced to be non-empty string, as we're sure they are always valued when coming from db.
-export type ResourceT = t.TypeOf<typeof ResourceT>;
+export type CosmosResource = t.TypeOf<typeof CosmosResource>;
```
gunzip, line 35: why renaming?

src/models/__tests__/user_data_processing.test.ts
```diff
@@ -120,9 +124,7 @@ describe("createOrUpdateByNewOne", () => {
 
     const model = new UserDataProcessingModel(containerMock);
 
-    const result = await model
-      .createOrUpdateByNewOne(aUserDataProcessing)
-      .run();
+    const result = await model.upsert(aNewUserDataProcessing).run();
```
gunzip, line 127: why this change here ? it looks like [createOrUpdateByNewOne](https://github.com/pagopa/io-functions-commons/blob/47a80c858b77fef036b27efd209c66b952e67440/src/models/user_data_processing.ts#L110) does more than a simple upsert

cloudify:
This was actually my first design choice, bue later I've decided to not add this requirement to the Cosmos model as it may be left as a decision for each concrete implementation of the models - what benefit do you see in making this mandatory?

src/utils/conversions.ts
balanza: I found this methods not to be used. Moreover, the linter now asks me to strip manual casting off, so what these methods are is actually just `as ModelId`

src/utils/messages.ts
balanza: debugging stuff, I remove it

gunzip:
it looks like it needs rebase

PR #85, opened 2020-07-31
Review - Add AsyncIterator / Iterable map, reduces, flatten 

balanza:
Following the conversation on #82, I'm adding some refinement on the `async` module.


src/utils/async.ts
```diff
@@ -87,9 +88,9 @@ export const filterAsyncIterable = <T, K = T>(
  * @param iter Original AsyncIterator
  * @param predicate Predicate function
  */
-export function filterAsyncIterator<T, K = T>(
-  iter: AsyncIterator<T | K>,
-  predicate: (value: T | K) => value is K
+export function filterAsyncIterator<T, K extends T>(
+  iter: AsyncIterator<T>,
+  predicate: (value: T) => value is K
```
balanza, line 93: @BurnedMarshal this change allow the following:
```ts
const iterator: AsyncIterator<string | number> = /* my iterator definition */

// type is correctly inferred as AsyncIterator<string>
const filteredIterator = filterAsyncIterator(iterator, (e): e is string => typeof e === "string");
```

BurnedMarshal: I like it :D

gunzip:
it looks like this needs a rebase

src/utils/async.ts
cloudify: can we also have a non type guard definition? like https://github.com/microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L1182

balanza: I think we can support both behaviours:
```ts
type Generic = { foo: string | number }
type Specialized = { foo: string }

const iterator: AsyncIterator<Generic> = /* implemetation */;

// with guards
// the tipe of the AsyncIterator is refined according to the type of the guard
const filtered1: AsyncIterator<Specialized> = filterAsyncIterator(iterator, (e): e is Specialized => typeof e.foo === "string")

// without guards
// the tipe of the AsyncIterator IS NOT refined according to the type of the guard, it's the same of the original iterator
const filtered2: AsyncIterator<Generic> = filterAsyncIterator(iterator, e => e==="foobar")
```

I think we should try to use the former, at the cost of having to define a specialized type.
The latter is a convenience pattern to get quick results. 


```

Comment (BurnedMarshal) [reply src/utils/async.ts:67]: You could use Function Overloading.
```typescript
export function filterAsyncIterator<T, K extends T>(
  iter: AsyncIterator<T>,
  predicate: (value: T) => value is K
): AsyncIterator<K>;

export function filterAsyncIterator<T>(
  iter: AsyncIterator<T>,
  predicate: (value: T) => boolean
): AsyncIterator<T>;

export function filterAsyncIterator<T>(
  iter: AsyncIterator<T>,
  predicate: (value: T) => boolean
): AsyncIterator<T> { // Implementation
```

balanza: that's the idea, let's see if it works

PR #87, opened 2020-07-31
More on AsyncIterator / Iterable

balanza:
Continuing the work on #85,  I implemented the following:
* `filterAsyncIterator` can be used with both guards or simple boolean unary function (read [here](https://github.com/pagopa/io-functions-commons/pull/85#discussion_r463454936) the rationale) 
* ~`mapAsyncIterator` can map the return value, if requested.~. **REVERTED**

~About the latter, the point is that `AsyncIterator`'s definition is inconsistent in term of whether `value` is valued or not when `done=true`. We have to specify if we want the mapping function to be applied to the returned value, otherwise we might be faced an undefined value. So I added an optional `mapResult` flag to enable this feature. I chose to have it disabled by default, as it's how our models' implementation will expect to be used.~


src/utils/async.ts
```diff
@@ -1,16 +1,39 @@
...
+  mapReturn: boolean | undefined
 ): AsyncIterator<V> {
   return {
     next: () =>
-      iter.next().then((result: IteratorResult<T>) =>
-        // IteratorResult defines that when done=true, then value=undefined
-        // that is, when the iterator is done there is no value to be procesed
-        result.done ? result : { done: false, value: f(result.value) }
+      iter.next().then((result: IteratorResult<T, TReturn>) =>
+        // IteratorResult defines that when done=true, then value is of a different kinf (TReturn instead of T)
```
gunzip, line 30:
```suggestion
        // IteratorResult defines that when done=true, then value is of a different kind (TReturn instead of T)
```

src/utils/async.ts
```diff
@@ -1,16 +1,39 @@
...
+ * @param f the functor
+ * @param mapReturn true if the return value must be mapped. Default: false
+ *
+ * @returns an aync iterator which values are mapped
  */
 export function mapAsyncIterator<T, V>(
   iter: AsyncIterator<T>,
-  f: (t: T) => V
+  f: (t: T) => V,
+  mapReturn?: false
```
gunzip, line 12: IMHO there's no need to map the return value. The map / filter semantic should follow the behavior of a `for of` loop (which stop reading before done is true)

balanza: I agree, but it's actually legal to write this
```ts
function*(){
    yield 1;
    yield 2;
    return 3;
}
```

I'm just afraid we'll encounter cases in which we fail using map because of this

gunzip: that would be a bad usage of the iterator, I'd go for the simpler solution by now

PR #88, opened 2020-08-04
cosmosDB model bugfix

AleDore:
This PR contains:
- `getNextVersion` bug fix on versioned models that have partitionKey field different from ModelId field
- A binding error fix on `findAllByQuery` method
- Improvement of `createOrUpdateByNewOne` in order to manage correctly upsert operation
- Some integration tests

### DIFF
--- a/src/utils/cosmosdb_model_versioned.ts
+++ b/src/utils/cosmosdb_model_versioned.ts
@@ -126,11 +126,7 @@ export abstract class CosmosdbModelVersioned<
     // if we get an explicit version number from the new document we use that,
     // or else we get the last version by querying the database
     const currentVersion: NonNegativeInteger | undefined = o.version;
-    // tslint:disable-next-line: no-console
-    console.log(`currentVersion => ${currentVersion}`);
     const modelId = this.getModelId(o);
-    // tslint:disable-next-line: no-console
-    console.log(`modelId => ${modelId}`);
     return (currentVersion === undefined
       ? this.getNextVersion(modelId, partitionKey)
       : fromEither<CosmosErrors, NonNegativeInteger>(right(currentVersion))
--- a/src/utils/cosmosdb_model_versioned.ts
+++ b/src/utils/cosmosdb_model_versioned.ts
@@ -66,7 +66,7 @@ export function generateVersionedModelId(
   return `${modelId}-${paddedVersion}`;
 }
 
-const incVersion = (version: NonNegativeInteger) =>
+export const incVersion = (version: NonNegativeInteger) =>
   (Number(version) + 1) as NonNegativeInteger;
 
 /**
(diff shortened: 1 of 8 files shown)


src/utils/cosmosdb_model_versioned.ts
```diff
@@ -118,20 +118,31 @@ export abstract class CosmosdbModelVersioned<
...
+  public upsert = (
+    o: TN,
+    requestOptions?: RequestOptions,
+    partitionKey?: string
+  ): TaskEither<CosmosErrors, TR> => {
     // if we get an explicit version number from the new document we use that,
     // or else we get the last version by querying the database
     const currentVersion: NonNegativeInteger | undefined = o.version;
+    // tslint:disable-next-line: no-console
+    console.log(`currentVersion => ${currentVersion}`);
```
balanza, line 130:
```suggestion
```

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -118,20 +118,31 @@ export abstract class CosmosdbModelVersioned<
...
     const modelId = this.getModelId(o);
+    // tslint:disable-next-line: no-console
+    console.log(`modelId => ${modelId}`);
```
balanza, line 133:
```suggestion
```

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -66,7 +66,7 @@ export function generateVersionedModelId(
   return `${modelId}-${paddedVersion}`;
 }
 
-const incVersion = (version: NonNegativeInteger) =>
+export const incVersion = (version: NonNegativeInteger) =>
```
balanza, line 69: Why do we need to export this?

balanza, line 69: Why do we need to export this?

src/utils/cosmosdb_model_versioned.ts
balanza, line 130:
```suggestion
```

src/utils/cosmosdb_model_versioned.ts
balanza, line 133:
```suggestion
```

src/utils/cosmosdb_model_versioned.ts
AleDore: It can be useful when we have to manually set next version on client side (i.e: during an upsert)

cloudify: that's the default behavior though, if the client just wants to increment the version it just have to avoid passing it along

gunzip: that's right but in this case the upsert will make another query to get the lastest version. in some cases is desiderable to get the latest version (in the caller) then try to update that object incrementing the version outside the upsert (saving a further query) in order to make on optimistic lock

balanza: Afaik the following will fail, right? 
```ts
const baseItem = { modelId: "AAA" }
const item0 = upsert(baseItem) // { id: "AAA-000",  modelId: "AAA", version: 0 }
const item1 = upsert(item0) // FAILS for 409 conflict - cannot insert id: "AAA-000"
```
I'm just not sure we have this very behaviour in the previous version.

gunzip: Yep, but the latter is more correct (upsert returns the upserted record with its version)

AleDore: > Afaik the following will fail, right?
> 
> ```ts
> // pseudocode
> const baseItem = { modelId: "AAA" }
> const item0 = upsert(baseItem) // OK { id: "AAA-000",  modelId: "AAA", version: 0 }
> const item1 = upsert(item0) // FAIL for 409 conflict - cannot insert id: "AAA-000"
> ```
> 
> I'm just not sure we have this very behaviour in the previous version.

Yes item0's upsert will fail, or even you have to set version to nextVersion.

balanza: For me it's ok, but it's a different behaviour than we had before refactor all models. See here: https://github.com/pagopa/io-functions-commons/blob/v10.7.0/src/utils/documentdb_model_versioned.ts#L109
tl;dr document version was always ignored.

We need to review all usages 🤦

gunzip: yes we do! luckyly we have a few upsert for versioned models (services, profiles and user-data-processing)

balanza: I made a quick search:
* https://github.com/pagopa/io-functions-admin/blob/a1bb2b16edff9c04bd86da6cffe405b097ddacaf/SetUserDataProcessingStatusActivity/handler.ts#L113
* https://github.com/pagopa/io-functions-admin/blob/4dc527550c001b2b9d8c2137effa1bac39c30dd0/UpdateService/handler.ts#L93

It seems it's just those 2

PR #90, opened 2020-08-06
[#174189493] DocumentSearchKey

balanza:
Introducing the concept of `DocumentSearchKey` which defines a tuple for safely searching a document on Cosmosdb. The tuple is `[modelId, partitionKey]` or, in case they refer the very same document field, just `[modelId]`

### Motivation
When searching a document using a model, it's not clear if and what partition key must be provided. This information depends on the structure of the related Cosmosdb collection, which is very far from the actual model usage. This PR is an attempt to encode such information into the type system, and try to make invalid query unrepresentable in the app.

### Breaking changes
* **`CosmosdbModel`** now accepts an optional type parameter `PARTITION_KEY = undefined | keyof TR` which represent the literal type for the partitionkey field name. When provided, the `DocumentSearchKey` associated with this model will be the pair `[modelId, partitionKey]`
* **`CosmosdbModel.find`** parameters are now `(searchKey: DocumentSearchKey<TR, MODEL_ID_KEY, PARTITION_KEY>, options?: RequestOptions)`. 
Before they were `(documentId: string. partitionKey: string, options?: RequestOptions)`.
* **`CosmosdbModelVersioned`** now accepts a required type parameter `MODEL_ID_KEY = keyof T` as well as an optional type parameter `PARTITION_KEY = undefined | keyof TR` . They represent respectively the model id and the partitionkey field names. When `MODEL_ID_KEY` is provided, the `DocumentSearchKey` associated with this model will be the pair `[modelId, partitionKey]`
* **`CosmosdbModelVersioned`** **constructor** accepts an optional parameter of type `PARTITION_KEY`, which is required when `PARTITION_KEY` type parameter is provided.
* **`CosmosdbModelVersioned.upsert**` parameters are now `(o: TN, requestOptions?: RequestOptions)`. Before they were `(o: TN, requestOptions?: RequestOptions, partitionKey?: string)`
* **`CosmosdbModelVersioned.findLastVersionByModelId`** parameters are now `(searchKey: DocumentSearchKey<T, MODEL_ID_KEY, PARTITION_KEY>)`. Before they were `(modelId: string, partitionKey?: string)`

The above changes impact the following model implementations (see [here](https://github.com/pagopa/io-functions-commons/commit/1d933b037c038d785fa4f6a3c6021df1ca2de120)):
* `MessageModel`
* `MessageStatusModel`
* `NotificationModel`
* `NotificationStatusModel`
* `ProfileModel`
* `ServiceModel`
* `UserDataProcessingModel`
 
### Other changes
* `CosmosdbModelVersioned` implements a protected method ` getSearchKey(document: T): DocumentSearchKey<T, MODEL_ID_KEY, PARTITION_KEY>` which automatically extracts the search key tuple from a given document. This is used internally by `CosmosdbModelVersioned.upsert`


src/utils/cosmosdb_model.ts
```diff
@@ -30,6 +29,30 @@ export const BaseModel = t.interface({
 
 export type BaseModel = t.TypeOf<typeof BaseModel>;
 
+/**
+ * Model a tuple which defines the references to search for a document.
+ * A Cosmodb document must be looked-up by its Identity alongside its PartitionKey. If PartitionKey field is the same of Identity field, it can be omitted.
+ * Hence this type models both cases: (ID) or (ID,PK) respectively if PK literal type is omitted or provided.
+ * @param T the type of the document mapped by the model
+ * @param ID the literal type defining the ID field for the document.
+ * @param PK (optional) the literal type defining the eventual partition key for the document. Default: undefined
```
gunzip, line 38:
```suggestion
 * @param PK (optional) the literal type defining the name of the partition key field. Default: undefined
```

src/utils/cosmosdb_model.ts
```diff
@@ -30,6 +29,30 @@ export const BaseModel = t.interface({
...
+ */
+export type DocumentSearchKey<
+  // T might not include base model fields ("id"), but we know they are mandatory in cosmos documents
+  // Quick win would be to narrow T to extend BaseModel, but that way we'd lose usage flexibility
```
gunzip, line 42: what do you mean with " we'd lose usage flexibility" ?

src/utils/cosmosdb_model.ts
```diff
@@ -30,6 +29,30 @@ export const BaseModel = t.interface({
...
+  // Hence we omit "extends BaseModel", but we check keys to be part of "T & BaseModel"
+  T,
+  ID extends keyof (T & BaseModel),
+  PK extends undefined | keyof (T & BaseModel) = undefined
+> = (T & BaseModel)[ID] extends string | number // narrow type to the ones that might be an identity
```
gunzip, line 47: in which case pk may be a number? (the literal type of pk field name)

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -75,13 +76,16 @@ export const incVersion = (version: NonNegativeInteger) =>
 export abstract class CosmosdbModelVersioned<
   T,
   TN extends Readonly<T & Partial<NewVersionedModel>>,
-  TR extends Readonly<T & RetrievedVersionedModel>
-> extends CosmosdbModel<T, TN & BaseModel, TR> {
+  TR extends Readonly<T & RetrievedVersionedModel>,
+  IDENTITY extends keyof T,
```
gunzip, line 80: if this is the modelId can e name it 'ModelId' ?

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -75,13 +76,16 @@ export const incVersion = (version: NonNegativeInteger) =>
...
+  PK extends keyof T | undefined = undefined
```
gunzip, line 81: I'd name this PartitionKey since PK might be "Primary Key" (which is not)

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -149,9 +152,9 @@ export abstract class CosmosdbModelVersioned<
    *  to avoid multi-partition queries.
    */
   public findLastVersionByModelId(
-    modelId: string,
-    partitionKey?: string
+    searchKey: DocumentSearchKey<T, IDENTITY, PK>
```
cloudify, line 155: I like the idea but I would like to keep the two params separated, I don't think the array is actually needed:

```ts
public findLastVersionByModelId(
  modelId: DocumentSearchKey<T, IDENTITY, PK>[0],
  partitionKey: DocumentSearchKey<T, IDENTITY, PK>[1]
)
```

gunzip: since partitionKey may be undefined it would be an optional parameter which in part defeat the purpose of this PR which is to use only required parameters once the model is defined

cloudify: I see that this must be a single object

src/utils/cosmosdb_model.ts
cloudify: I believe the PK is always a `string`

cloudify:
From a quick read, I can see how this improves the type safety of our code, I can't say I 100% understand it but it looks good to me :)

src/utils/cosmosdb_model.ts
balanza: tl;dr you can pass `MyModel` and not necessarily `RetrievedMyModel`

`T` is the struct which define our model schema. It's not required to have a `id` field. Then we have some type definitions that extend (or _compose with_) `T` and they do have such field, but they don't overlap each other. So you would have needed to oveload `DocumentSearchKey` or any child type - a little mess.
I thought that the easier interface was to use the minimum subset of definition (that is: `T`) and then just make the constrant that `ID` and `PK` must fall in the set `keyof T | "id"`.

I hope that makes sense ;)

src/utils/cosmosdb_model.ts
balanza: true

src/utils/cosmosdb_model_versioned.ts
balanza: well, Modeld stands for Model Identity, and we are in the context of a model, so....

I can rename it, but I'll also have to deal with `ModelId` which is a decoder that maps an id field value (just a string decoder so far).

src/utils/cosmosdb_model_versioned.ts
balanza: ok

src/utils/cosmosdb_model_versioned.ts
balanza: If you split parameters you will have a signature like this:
```ts
find(
  modelId: DocumentSearchKey<T, IDENTITY, PK>[0],
  partitionKey: DocumentSearchKey<T, IDENTITY, PK>[1],
  options?: RequestOptions
)
```

Which leads to usages like
```ts
find("abc") // ERROR!
find("abc", undefined) // OK
find("abc", undefined, { /* options */ }) // OK
```

----

> I see that this must be a single object

Do you mean something like this? 
```ts
find({
      modelId,
      partitionKey
  }: DocumentSearchKey<T, IDENTITY, PK>,
  options?: RequestOptions
)
```

I think this leads to a more verbose api
```ts
find({ modelId: "abc", partitionKey: "def" })
```

cloudify: I mean, after I proposed to split the tuple to two params, I checked the rest of the code and realized why you kept it as a single object/tuple

balanza: What I can do is to consider a scalar instead of a tuple in the case of no partition key
```ts
model.find("abc") // OK
model.find(["abc"]) // ERROR
partitionedModel.find(["abc", "def"]) // OK
partitionedModel.find(["abc"]) // ERROR
partitionedModel.find("abc") // ERROR
```

It's a sugar that isn't worth the effort IMO

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -75,13 +72,16 @@ export const incVersion = (version: NonNegativeInteger) =>
 export abstract class CosmosdbModelVersioned<
   T,
   TN extends Readonly<T & Partial<NewVersionedModel>>,
-  TR extends Readonly<T & RetrievedVersionedModel>
-> extends CosmosdbModel<T, TN & BaseModel, TR> {
+  TR extends Readonly<T & RetrievedVersionedModel>,
+  MODEL_ID_KEY extends keyof T,
```
gunzip, line 76: nit: it looks like by convention we are using upper cases types only for acronyms or single characters (ie. `T`, `TN`) while for extended names (ie. `BaseModel`) we're using PascalCase so i guess these may be better named as
`ModelIdKey` and `PartitionKey` (or `ModelIdKeyT`, `PartitionKeyT` in case of conflicts)

balanza: correct, and it should have been obvious due the fact that we have
```ts
    protected readonly modelIdKey: ModelIdKey,
    protected readonly partitionKey?: PartitionKey
```

src/utils/cosmosdb_model.ts
```diff
@@ -173,14 +197,16 @@ export abstract class CosmosdbModel<
...
    * Retrieves a document from the document ID.
    *
-   * @param documentId    The ID of the document to retrieve.
+   * @param searchKey    The tuple of values used to look-up a document. It can be [documentId] or [documentId, partitionKey] depending on the definition of the model instance.
    * @param partitionKey  The partitionKey associated to this model.
    */
   public find(
-    documentId: string,
-    partitionKey: string,
+    searchKey: DocumentSearchKey<TR, IDENTITY, PartitionKey>,
```
gunzip, line 204: IDENTITY = ModelIdKey ?

gunzip:
it looks like this needs a rebase

PR #92, opened 2020-08-07
Create ServicePayload definition

francescopersico:
Split Service definition in:

ServicePayload: used to create/update service in services function
Service: used as return type in get/create/update in services function, used as create/update type in admin function

### DIFF
--- a/openapi/definitions.yaml
+++ b/openapi/definitions.yaml
@@ -527,18 +527,18 @@ SenderMetadata:
     - service_name
     - organization_name
     - department_name
-Service:
-  description: A Service tied to an user's subscription.
+ServicePayload:
+  description: A payload used to create/update a service for a user.
   type: object
   properties:
-    service_id:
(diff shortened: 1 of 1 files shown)


openapi/definitions.yaml
```diff
@@ -527,18 +527,18 @@ SenderMetadata:
     - service_name
     - organization_name
     - department_name
-Service:
-  description: A Service tied to an user's subscription.
+ServicePayload:
+  description: A payload used to create/update a service for a user.
```
gunzip, line 531: why remove service_id ? (we are used to create a service provinding its id)

francescopersico: If needed you can use the Service type

gunzip:
is this mergeable? lgtm

AleDore (commented):
lgtm. `Service` definition is unchanged.

PR #97, opened 2020-08-11
[#174189493] Review CosmosdbModelVersioned public interface

balanza:
This introduces new methods and breaking changes to old ones for the purpose of handling different scenarios explicitly when saving versioned documents.

### Changes
* ~type `NewVersionedModel` doesn't include `version` field anymore.~
* type `NewVersionedModel` has been removed. `TN` is actually `T`, but it can be extended (to add tags for example).
* `create(o:TN)` creates a new document on db which version is always `0`.
* `upsert(o:TN)` creates a new document on db which version is calculated from the latest document present on db for the same modeld (`0` if no documents are found)
* `update(o: TR)` creates a new document on db which version is calculated as `o.version + 1`

### DIFF
--- a/src/utils/__tests__/cosmosdb_model_versioned.test.ts
+++ b/src/utils/__tests__/cosmosdb_model_versioned.test.ts
@@ -16,6 +16,7 @@ import { NonEmptyString } from "italia-ts-commons/lib/strings";
 import { BaseModel } from "../cosmosdb_model";
 import {
   CosmosdbModelVersioned,
+  generateVersionedModelId,
   NewVersionedModel,
   RetrievedVersionedModel
 } from "../cosmosdb_model_versioned";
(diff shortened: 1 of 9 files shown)


src/utils/cosmosdb_model_versioned.ts
```diff
@@ -223,4 +207,35 @@ export abstract class CosmosdbModelVersioned<
...
+    );
+  }
+
+  /**
+   * Strips off meta fields which are nor part of the base model definition
+   * @param o
+   */
+  private toBaseType(o: TR): T {
+    const { _etag, _rid, _self, _ts, id, version, ...n } = o;
+    return (n as unknown) as T;
```
BurnedMarshal, line 239: Can have Type `T` an `id` field?

balanza: why?

gunzip, line 239: cn we use decode instead of casting?

BurnedMarshal: This method removes the `id` field from the TR. But maybe `id` is reserved in a versioned model.

src/utils/__tests__/cosmosdb_model_versioned.test.ts
```diff
@@ -124,135 +128,130 @@ const errorResponse: ErrorResponse = new Error();
...
-    const expectedNextVersion = 3;
-
+    const expectedNextVersion = 0;
+    const modelId = aNewMyDocument[aModelIdField];
     const model = new MyModel(container);
 
-    const result = await model
-      .upsert({ ...aNewMyDocument, version: modelCurrentVersion })
-      .run();
+    const result = await model.create(aNewMyDocument).run();
```
gunzip, line 207: this needs to be changed

src/utils/cosmosdb_model_versioned.ts
balanza: In versioned models `id` is a meta-data field, as it doesn't ship data and it's basically a Cosmos thing. I don't like the manual cast but I see no other way to skim unwanted fields.

PR #112, opened 2020-10-09
[#175168739] Common utilities for mailing

balanza:
Collect mail utilities and groups them under a unique submodule of `io-functons-commons`. The idea is to have a single, shared entrypoint applications can call and setup by just pour their configuration in it.

The PR is in draft as I'd like to make some integration tests.

### Design
Everything regarding email is now under the `/mailer` folder. There are no outbound dependency references, so the folder may be even moved in a module on its own.
The folder has a `index.ts` entrypoint which is ideally the only file users want to refer. Inner references worth to the user are proxied by the index module itself.
Every reference to `nodemailer` has been confined into the `mailer/transports` file.
Mailer's env configuration decoder is exposed and user applications may want to extend their config with that.

### Usage
```ts
import { sendMail, getMailerTransporter, MailerConfig } from "io-functions-commons/dist/mailer";

// config is expected to be decoded already
const config = MailerConfig.decode(process.env).getOrElseL(_=> throw);
// create a transporter
const transporter = getMailerTransporter(config);
// use sendMail util to wrap the request into a TaskEither
sendMail(transporter, { from, to, subject /* etc etc */ }).run();
```

### Breaking changes
When upgrading to a version of `io-functions-commons` which introduces this changes, please take note of the following:
* Transporter options [used](https://github.com/pagopa/io-functions-app/blob/6115472a8ecc5e3451befb22efa05ab327fca2ae/SendValidationEmailActivity/index.ts#L61) [to](https://github.com/pagopa/io-functions-admin/blob/b4b1c8da1ccdb3cfbcf12bdc23b3e318681fb7ce/SendUserDataDeleteEmailActivity/index.ts#L33) [be](https://github.com/pagopa/io-functions-services/blob/a1e62f943e6e0c7cbaade4fdd8c3c40c7b2ad9f2/EmailNotificationActivity/index.ts#L84) composed inside the application, now it's `getMailerTransporter`'s duty
* module `utils/nodemailer` is removed, most of its functionalities are now in `mailer/transports`
* modules `utils/email` is removed, its only function (`sendMail`) has been included into `mailer/transports`
* function `sendMail` used to return a `Promise<Either<Error, MailInfo>>`, now it's a `TaskEither<Error, MailInfo>`
* modules `utils/multi_transport_connection` and `utils/mailup` have both been moved into `/mailer`


package.json
```diff
@@ -23,7 +23,7 @@
     "@types/helmet": "^0.0.40",
     "@types/jest": "^24.0.13",
     "@types/node": "10.14.1",
-    "@types/nodemailer": "^4.6.2",
+    "@types/nodemailer": "^6.4.0",
```
balanza, line 26: This is fine as we are already using this version in our applications

src/mailer/index.ts
```diff
@@ -0,0 +1,106 @@
...
+// 5 seconds timeout by default
+const DEFAULT_EMAIL_REQUEST_TIMEOUT_MS = 5000;
+
+// Must be an https endpoint so we use an https agent
+const abortableFetch = AbortableFetch(agent.getHttpsFetch(process.env));
+const fetchWithTimeout = setFetchTimeout(
+  DEFAULT_EMAIL_REQUEST_TIMEOUT_MS as Millisecond,
+  abortableFetch
+);
+const fetchAgent = toFetch(fetchWithTimeout);
```
gunzip, line 48: I think, since this is accessing process.env, this should be configured in the caller and removed from this module but passed as a paramtere to `getMailerTransporter`

src/mailer/config.ts
```diff
@@ -0,0 +1,84 @@
+import * as t from "io-ts";
+import { NonEmptyString } from "italia-ts-commons/lib/strings";
+import { MailMultiTransportConnectionsFromString } from "./multi_transport_connection";
+
+// exclude a specific value from a type
+// as strict equality is performed, allowed input types are constrained to be values not references (object, arrays, etc)
+// tslint:disable-next-line max-union-size
+export const AnyBut = <A extends string | number | boolean | symbol, O = A>(
```
gunzip, line 8: this belongs to io-ts-commons, but at least, can we move it into ustils/types.ts ?

gunzip (changes requested):

src/mailer/config.ts
balanza: I'd rather not do it because I didn't stress `AnyBut` enough to make it a general-purpose utility. It is performing good for this specific case, but I'm afraid it might not be the same in other contexts.

gunzip:
why draft?

gunzip (approved):

PR #113, opened 2020-10-12
[#175225851] Mailer integration tests

balanza:
This PR points toward #112 as we might ship the current changes along with #112's. However, I kept it separated for having a dedicated discussion.

Changes introduce some integration tests for the `mailer` submodule. This PR actually works on 2 aspects:

### 1. Design an integration test environment
The following changes are introduced to serve all future integration tests, hence they require dedicated analysis:
* a dedicated `__integrations__` folder which contains a separated project to run integration tests (let's decide the name of this folder)
* a step in the pipeline
* a `test:integration` command in the root's `package.json`

### 2. Test suite for the mailer
* We can test only Mailhog so far, let's see if we can test Sendgrid and MailUp as well (maybe with test account, is it feasible @gunzip?)
* I'd like to test the received messages, ~but I found no way to connect to Mailhog via imap/pop3~ use Mailhog api


gunzip:
1) we don't have test accounts but i think one dev could create one (but anyway we cannot make tests run into our pipeline). Anyway the only thing you could test is if you get 200OK calling the mailup or sendgrid API (very little indeed)
2) mailhog as an api to do that https://github.com/mailhog/MailHog/blob/master/docs/APIv2/swagger-2.0.yaml

__integrations__/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "io-functions-commons-integrations",
+  "version": "1.0.0",
+  "description": "Inegration test suite for integration testing",
```
AleDore, line 4:
```suggestion
  "description": "Integration test suite for integration testing",
```

package.json
```diff
@@ -14,6 +14,7 @@
     "build": "yarn generate:definitions && tsc",
     "postversion": "git push && git push --tags",
     "test": "jest -i",
+    "test:integration": "yarn build && cd ./__integrations__ && yarn install --frozen-lockfile && yarn start",
```
AleDore, line 17: I'd keep add a `test:all` script that performs unit and integration tests.

balanza:
> but anyway we cannot make tests run into our pipeline

@gunzip what do you mean?

package.json
balanza: what about I add a script 
```json
"test:all": yarn test && yarn test:integration
```
?

AleDore: Yes 👍

PR #114, opened 2020-10-29
[#175160851] Add VisibleServicePayload and HiddenServicePayload definitions

infantesimone:
This PR introduces a new definition for `ServicePayload` in order to model the mandatory of service metadata for visible services.

So, the new `ServicePayload`  is valid against one of the following schema:

 - `VisibleServicePayload` 
 - `HiddenServicePayload`

These definitions differ only for the metadata property, which in the case of `VisibleServicePayload` is required.

### DIFF
--- a/openapi/definitions.yaml
+++ b/openapi/definitions.yaml
@@ -532,7 +532,7 @@ ServicePayload:
   x-one-of: true
   allOf:
    - $ref: "#/VisibleServicePayload"    
-   - $ref: "#/InvisibleServicePayload"
+   - $ref: "#/HiddenServicePayload"
 CommonServicePayload:
   description: Common properties for a ServicePayload
   type: object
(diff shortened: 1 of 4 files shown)


openapi/definitions.yaml
```diff
@@ -553,19 +558,41 @@ ServicePayload:
...
     - authorized_cidrs
+VisibleServicePayload:
+  description: A payload used to create/update a service that appears in the service list.
+  allOf:
+  - $ref: "#/CommonServicePayload"
+  - type: object
+    properties:
+      is_visible:
+        type: boolean
+        enum: [ true ]
```
gunzip, line 575: why not `default: true `?

infantesimone:
`oneOf` is not support from the swagger version 2.0

gunzip:
you can emulate it in italia-utils adding x-one-of

https://github.com/pagopa/io-utils/blob/1d05f9ca244bb628dc8123d4d3f4fbd44423e3e1/__mocks__/api.yaml#L167

AleDore (commented):
Why draft?

infantesimone:
io-utils not support boolean with enum require from `VisibleServicePayload` and `InvisibleServicePayload` definitions, so after [this](https://github.com/pagopa/io-utils/pull/208)  I can conclude this PR.

gunzip (approved):
@balanza can we merge this?

balanza (changes requested):
As we're touching definitions being consumed by other projects, I'd add some tests to be sure we're not introducing behavioural changes

balanza (approved):
very good 👍

PR #117, opened 2020-11-11
[#169589538] Fix mjml raw template styling

AleDore:
A minimal fix to generated style on email template.

### DIFF
--- a/src/templates/html/default.ts
+++ b/src/templates/html/default.ts
@@ -33,17 +33,9 @@ export default function(
       }</style><style type="text/css">@media only screen and (max-width:480px) {
       table.mj-full-width-mobile { width: 100% !important; }
       td.mj-full-width-mobile { width: auto !important; }
-    }</style><style type="text/css">#outlook a { padding:0; }
-          .ReadMsgBody { width:100%; }
-          .ExternalClass { width:100%; }
-          .ExternalClass * { line-height:100%; }
-          body { margin:0;padding:0;-webkit-text-size-adjust:100%;-ms-text-size-adjust:100%; }
-          table, td { border-collapse:collapse;mso-table-lspace:0pt;mso-table-rspace:0pt; }
-          img { border:0;height:auto;line-height:100%; outline:none;text-decoration:none;-ms-interpolation-mode:bicubic; }
-          p { display:block;margin:13px 0; }</style><!--[if !mso]><!--><style type="text/css">@media only screen and (max-width:480px) {
-            @-ms-viewport { width:320px; }
-            @viewport { width:320px; }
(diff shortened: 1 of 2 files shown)


gunzip:
can you please add a screenshot with before and after?

src/templates/html/default.ts
```diff
@@ -33,7 +33,17 @@ export default function(
       }</style><style type="text/css">@media only screen and (max-width:480px) {
       table.mj-full-width-mobile { width: 100% !important; }
       td.mj-full-width-mobile { width: auto !important; }
-    }</style></head><body style="background-color:#FFFFFF;"><div style="background-color:#FFFFFF;"><!-- Header --><!-- <mj-section padding="10px 0">
+    }</style><style type="text/css">#outlook a { padding:0; }
```
gunzip, line 36: I guess we should use this instead:

https://documentation.mjml.io/#mj-style

templates/mjml/default.mjml
```diff
@@ -1,12 +1,27 @@
 <mjml>
   <mj-head>
+    <mj-raw>
```
gunzip, line 3: why not https://documentation.mjml.io/#mj-style ?

templates/mjml/default.mjml
```diff
@@ -1,12 +1,27 @@
...
+      <style type="text/css">#outlook a { padding:0; }
+          .ReadMsgBody { width:100%; }
+          .ExternalClass { width:100%; }
+          .ExternalClass * { line-height:100%; }
```
gunzip, line 7: can you add some comments on what are these classes and where do they come from?

templates/mjml/default.mjml
```diff
@@ -1,12 +1,27 @@
...
+          body { margin:0;padding:0;-webkit-text-size-adjust:100%;-ms-text-size-adjust:100%; }
+          table, td { border-collapse:collapse;mso-table-lspace:0pt;mso-table-rspace:0pt; }
+          img { border:0;height:auto;line-height:100%; outline:none;text-decoration:none;-ms-interpolation-mode:bicubic; }
+          p { display:block;margin:13px 0; }</style><!--[if !mso]><!--><style type="text/css">@media only screen and (max-width:480px) {
+            @-ms-viewport { width:320px; }
+            @viewport { width:320px; }
```
gunzip, line 13: these should be already in the generated html, can you please check?

gunzip (changes requested):

gunzip (approved):

templates/mjml/default.mjml
AleDore: Yes it is present :)

PR #128, opened 2020-12-30
[#176013578] Remove prepare script

balanza:
`prepare` script is executed before and every time we run `npm publish`. We used to build the package in the `prepare` script, and this makes perfect sense when publishing from local environment.

As we moved the publication on the CI/CD environment, and we separated build and publish step, that is no longer needed.
As double check, I introduced a `prepack` script which basically confirms that we are shipping every needed file in the package.

The script is in bash: that introduces the constraint that we can only publish from environment that support such language. My assumption is that it's a reasonable restriction.

Plus: i removed `shx` as we no longer need it.


.devops/deploy-pipelines.yml
```diff
@@ -98,24 +98,13 @@ stages:
             inputs:
               SourceFolder: '$(System.DefaultWorkingDirectory)'
               TargetFolder: '$(System.DefaultWorkingDirectory)/bundle'
+              # The list of files to be considered is determined by the "files" entry in package.json
```
gquadrati, line 101: what do you mean?

package.json
```diff
@@ -9,7 +9,7 @@
     "dist/"
   ],
   "scripts": {
-    "prepare": "yarn build",
+    "prepack": "[[ ! -d ./dist ]] && echo \"Cannot create package: missing 'dist' folder\" && exit 1",
```
gquadrati, line 12: where this "prepack" script will be called?

.devops/deploy-pipelines.yml
gquadrati, line 101: what do you mean?

gquadrati (approved):
LGTM

package.json
balanza: Before `npm pack` is executed, and that includes `npm publish`. See https://docs.npmjs.com/cli/v6/using-npm/scripts#prepare-and-prepublish

.devops/deploy-pipelines.yml
balanza: this: https://github.com/pagopa/io-functions-commons/blob/ca9a943a9004bbf7aec7804c5715fbff80ab8fde/package.json#L8

gquadrati: yes, I saw it after this comment.. sorry about that!

package.json
gquadrati: thanks!

PR #135, opened 2021-03-12
[#177318059] Add constant common module

AleDore:
#### List of Changes
- Add constant module

#### Motivation and Context
Since `io-functions-cgn` and `io-functions-cgn-merchant` must share some constant values we can place them in io-functions-commons with a general purpose. The new constant module could provide a set of shared const that could suffice with our needs if we have to share non editable properties along functions.

See this [story](https://www.pivotaltracker.com/story/show/177318059) for further informations.

No tests needed

#### Types of changes

#### Checklist:

### DIFF
--- a/src/utils/constant.ts
+++ b/src/utils/constant.ts
@@ -1,0 +1,2 @@
+export const CGN_OTP_PREFIX = "OTP_";
+export const CGN_FISCALCODE_OTP_PREFIX = "OTP_FISCALCODE_";


balanza:
I don't think we should add references to CGN inside this module. `io-functions-commons` is already a bloated package, it has common utilities along with specific model definitions. 

I'm aware we never came out with a formal definition of what is a core domain and what is not in IO, so there might be fuzzy borders. Nevertheless, I'm confident CGN should not be here.

Things you could do instead: 
* convert such constants as environment configurations and define them in a shared terraform script
* a `io-functions-cgn-commons`, containing these constants plus shared models
* use a monorepo containing both projects, and share these constants locally

AleDore:
> I don't think we should add references to CGN inside this module. `io-functions-commons` is already a bloated package, it has common utilities along with specific model definitions.
> 
> I'm aware we never came out with a formal definition of what is a core domain and what is not in IO, so there might be fuzzy borders. Nevertheless, I'm confident CGN should not be here.
> 
> Things you could do instead:
> 
> * convert such constants into environment configurations and define them in a shared terraform script
> * a `io-functions-cgn-commons`, containing these constants plus shared models
> * use a monorepo containing both projects, and share these constants locally

cc @BurnedMarshal

BurnedMarshal:
> I don't think we should add references to CGN inside this module. `io-functions-commons` is already a bloated package, it has common utilities along with specific model definitions.
> 
> I'm aware we never came out with a formal definition of what is a core domain and what is not in IO, so there might be fuzzy borders. Nevertheless, I'm confident CGN should not be here.
> 
> Things you could do instead:
> 
> * convert such constants into environment configurations and define them in a shared terraform script
> * a `io-functions-cgn-commons`, containing these constants plus shared models
> * use a monorepo containing both projects, and share these constants locally



> > I don't think we should add references to CGN inside this module. `io-functions-commons` is already a bloated package, it has common utilities along with specific model definitions.
> > I'm aware we never came out with a formal definition of what is a core domain and what is not in IO, so there might be fuzzy borders. Nevertheless, I'm confident CGN should not be here.
> > Things you could do instead:
> > 
> > * convert such constants into environment configurations and define them in a shared terraform script
> > * a `io-functions-cgn-commons`, containing these constants plus shared models
> > * use a monorepo containing both projects, and share these constants locally
> 
> cc @BurnedMarshal

This is a good point. I would exclude the first option, because an environment configuration should be used for mutable values, and these particular constants should be immutable.
Maybe the second have a lesser integration cost in terms of deploy (pipeline) and integration test (io-mock).

balanza:
I'd say: just hardcode such values and write a meaningful comment will be enough

PR #136, opened 2021-03-22
[#ENTE-20] AuthorizedCIDRs without subnet

BurnedMarshal:
#### List of Changes
- `authorizedCIDRs` without subnet has default `\32` subnet.

#### Motivation and Context
When a Service has an authorized CIDR without a subnet the IPChek fail. This PR adds a default `\32` subnet.

- unit test

#### Types of changes

- [ ] Chore
- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/utils/source_ip_check.ts
+++ b/src/utils/source_ip_check.ts
@@ -16,6 +16,7 @@ import {
   ResponseErrorForbiddenNotAuthorized,
   ResponseErrorInternal
 } from "@pagopa/ts-commons/lib/responses";
+import { toAuthorizedCIDRs } from "../models/service";
 
 /**
  * Whether IP is contained in the provided CIDRs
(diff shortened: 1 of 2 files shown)


src/utils/source_ip_check.ts
```diff
@@ -162,5 +163,12 @@ export function clientIPAndCidrTuple(
   clientIp: ClientIp,
   userAttributes: IAzureUserAttributes
 ): ITuple2<ClientIp, ReadonlySet<string>> {
-  return Tuple2(clientIp, userAttributes.service.authorizedCIDRs);
+  return Tuple2(
+    clientIp,
+    toAuthorizedCIDRs(
+      Array.from(userAttributes.service.authorizedCIDRs).map(c =>
+        c.indexOf("/") !== -1 ? c : `${c}/32`
+      )
```
balanza, line 171: We should be more explicit about what's going on. We can either add a line of comment or write something like
```ts
const withDefaultSubnet = ip => ip.indexOf("/") !== -1 ? ip : `${ip}/32

[...]

Array.from(userAttributes.service.authorizedCIDRs).map(withDefaultSubnet)
```

balanza (dismissed):
LGTM apart minor style comment.

My only concern: that's indeed semantically correct, but are we going to introduce breaking changes with this?

src/utils/source_ip_check.ts
BurnedMarshal: fixed in 7211251

BurnedMarshal:
> LGTM apart minor style comment.
> 
> My only concern: that's indeed semantically correct, but are we going to introduce breaking changes with this?

No braking change. This is a bugfix because when a service has a CIDR without a subnet (is allowed by the type definition of [CIDR](https://github.com/pagopa/io-functions-commons/blob/master/openapi/definitions.yaml#L777)) the function `isContainedInCidrs` throw an exception. Using the new `clientIPAndCidrTuple` helper as extractor this error cannot happen.

balanza (approved):

PR #140, opened 2021-04-15
[#IP-82] Add a response header containing the current app version

fabriziopapi:
In order to ease the troubleshooting, we want to add to all the service response a new header field (X-API-Version) containing the current app version.
We created a middleware setting the response header and a common register function to call from all the service code.
Both those functions have been implemented in this project to match the DRY principle and to ease the future similar task.

### DIFF
--- a/src/utils/express.ts
+++ b/src/utils/express.ts
@@ -1,4 +1,5 @@
 import * as express from "express";
+import { toString } from "fp-ts/lib/function";
 import * as helmet from "helmet";
 import * as csp from "helmet-csp";
 import * as referrerPolicy from "referrer-policy";
@@ -39,4 +40,26 @@ export function secureExpressApp(app: express.Express): void {
       }
     })
   );
+}
+
+/**
+ * Create an express middleware to set the 'X-API-Version' response header field containing the current app version in execution (from the npm environment).
+ * @returns a factory method for the Middleware
+ */
+const createAppVersionHeaderHandler: () => express.RequestHandler = () => (
+  _,
+  res,
+  next
+) => {
+  res.setHeader("X-API-Version", toString(process.env.npm_package_version));
+  next();
+};
+
+/**
+ * Configure all the default express middleware handlers on the input express app.
+ * Register here all the non business-logic-related common behaviours.
+ * @param app an express application
+ */
+export function configureDefaultHandlers(app: express.Express): void {
+  app.use(createAppVersionHeaderHandler());
 }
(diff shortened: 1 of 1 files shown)


src/utils/express.ts
```diff
@@ -40,3 +41,29 @@ export const secureExpressApp = (app: express.Express): void => {
...
+ * Create an express middleware to set the 'X-API-Version' response header field to the current app version in execution (from the npm environment).
+ *
+ * @returns a factory method for the Middleware
+ */
+export const createAppVersionHeaderHandler: () => express.RequestHandler = () => (
+  _,
+  res,
+  next
+): void => {
+  fromNullable(process.env.npm_package_version).map(v =>
```
balanza, line 55: I cannot find that in docs, but I'm pretty sure `process.env.npm_package_version` is defined only when the application is started with `npm start` (or `yarn start`).
Although  I think that's the case of every app we make, it's a bold assumption. A safer approach would be to read `package.json` file 
```ts
import { version } from "../package.json";
```

src/utils/express.ts
```diff
@@ -40,3 +41,29 @@ export const secureExpressApp = (app: express.Express): void => {
...
+  next();
+};
+
+/**
+ * Configure all the default express middleware handlers on the input express app.
+ * Register here all the non business-logic-related common behaviours.
+ *
+ * @param app an express application
+ */
+export const configureDefaultHandlers = (app: express.Express): void => {
```
balanza, line 67: Naming: 

maybe _middlewares_ more than _handlers_?

balanza (changes requested):

src/utils/express.ts
fabriziopapi: Yes, the npm_package_version works only for npm-started app, but using the import approach delegate the "version" extraction to the middleware caller becouse it depends of the project structure ("../package.json").
We can easly take the verison as input in this function, attributing the version rightness responsability to the caller.
Do you think it's safer in this way?

src/utils/express.ts
fabriziopapi: Ok, naming convention fixed.

src/utils/express.ts
balanza: Good point, I didn't realize that we were going to do that from a package. 
We could pass the whole package to the factory method, or we can just read the file from 
```js
`${process.cwd()}/package.json`
```

Maybe the second option will provide the right level of abstraction. I don't have a strong opinion actually.

fabriziopapi: I've chained both the solution: the verison is extracted by package.json, if fails it will fallback to npm env version.

PR #142, opened 2021-04-23
[#IP-144] Added new optional property to UserDataProcessing model

michaeldisaro:
#### List of Changes
- Added optional 'reason' property to UserDataProcessing model
- Added branded type ValidUserDataProcessing to validate UserDataProcessing against status and reason property

#### Motivation and Context
We want to include a readable reason in UserDataProcessing when the processing fails so that we can figure out where it failed.

- yarn build
- yarn lint
- yarn test

#### Types of changes

#### Checklist:


BurnedMarshal (commented):
Can we use more strict typings?
```typescript
export const UserDataProcessing = t.intersection([
  CommonUserDataProcessing,
  t.union([
    t.interface({
      status: t.union([
        t.literal(UserDataProcessingStatusEnum.ABORTED),
        t.literal(UserDataProcessingStatusEnum.CLOSED),
        t.literal(UserDataProcessingStatusEnum.PENDING),
        t.literal(UserDataProcessingStatusEnum.WIP)
      ])
    }),
    t.interface({
      reason: NonEmptyString,
      status: t.literal(UserDataProcessingStatusEnum.FAILED)
    })
  ])
]);
```
Or we need to admit invalid `UserDataProcessing`?

michaeldisaro:
@BurnedMarshal yes, you're right! Thanks, your code fits better this case. I'll update it. By the way this PR should be a draft, I'm sorry for the inconvenient.

BurnedMarshal:
If we have documents into the database without `reason` we can use even `withDefault` to avoid decoding errors on reading.
```typescript
t.interface({
  reason: withDefault(NonEmptyString, "no reason found" as NonEmptyString),
  status: t.literal(UserDataProcessingStatusEnum.FAILED)
})
```

michaeldisaro:
I noticed that the above model is breaking the build, because it cause reason to be mandatory in some paths and is not excluding reason from other paths.

I tried to make a change locally:

export const UserDataProcessing = t.intersection([
  CommonUserDataProcessing,
  t.union([
    t.partial({
      status: t.union([
        t.literal(UserDataProcessingStatusEnum.ABORTED),
        t.literal(UserDataProcessingStatusEnum.CLOSED),
        t.literal(UserDataProcessingStatusEnum.PENDING),
        t.literal(UserDataProcessingStatusEnum.WIP)
      ]),
      reason: t.void
    }),
    t.interface({
      status: t.literal(UserDataProcessingStatusEnum.FAILED),
      reason: withDefault(NonEmptyString, "no reason found" as NonEmptyString)
    })
  ])
]);

The "reason: t.void" is excluding reason for those statuses, allowing for a right decoding without that field since it's a partial.

I'm having another problem... I'd like to talk with you tomorrow about this..

balanza:
In https://github.com/pagopa/io-functions-commons/pull/142/commits/865cdf9abc2235e3a3401f4357a9b650b7de6558 we introduced some counter-intuitive solutions to solve the problem. We are not happy with such implementations, still we have them to work and we consider them just fine for the moment.

I try to highlight some notable decisions for our future selves:
1. We had to **manually enumerate all possible status values** but `FAILED` in https://github.com/pagopa/io-functions-commons/blob/ip-144-user_data_processing-failure-reason-in-model/src/models/user_data_processing.ts#L81. This is a fragile solution as it does not auto-adapt in case the `UserDataProcessingStatusEnum` set of values changes. 
We'd rather use a codec like [`AnyBut`](https://github.com/pagopa/io-functions-commons/blob/0818badcdd2fe096fb7875119a3f15bb69a23a0e/src/mailer/config.ts#L8), thus considering _all values except `FAILED`_, but we couldn't make it work.
2. We had to define schema **without id field** first ([here](https://github.com/pagopa/io-functions-commons/blob/865cdf9abc2235e3a3401f4357a9b650b7de6558/src/models/user_data_processing.ts#L97)), then adding it at last ([here](https://github.com/pagopa/io-functions-commons/blob/865cdf9abc2235e3a3401f4357a9b650b7de6558/src/models/user_data_processing.ts#L105)). This has been necessary because `createOrUpdateByNewOne` method signature required such field not to be present. In former version of the module, such need used to be expressed by the built-in type `Omit`, letting Typescript to cope with that; as soon as we introduced the disjointed union for the status field, that somehow stopped working so we had to handle it manually.
3. Our intention was for the `reason` field to be defined only when `status=FAILED`, and not to be defined elsewhere. Instead, we had to have it defined **partial and void**. 
```ts
// what we intended 
const WithFailureReason = t.union([
  t.interface({
    status: t.union([ /* all but FAILED */ ])
  }),
  t.interface({
    reason: withDefault(NonEmptyString, "no reason found" as NonEmptyString),
    status: t.literal(UserDataProcessingStatusEnum.FAILED)
  })
]);

// what we have
const WithFailureReason = t.union([
  t.partial({
    reason: t.void,
    status: t.union([ /* all but FAILED */ ])
  }),
  t.interface({
    reason: withDefault(NonEmptyString, "no reason found" as NonEmptyString),
    status: t.literal(UserDataProcessingStatusEnum.FAILED)
  })
]);
```
That has been necessary to avoid `reason` field to leak into a document with status different than `FAILED`. Our first choice would have been to just cleanse such field using `t.exact` codec, but that somehow messed with meta-fields needed by CosmosDB models such as `RetrievedVersionedModel`.
```ts
// suppose an incoming object with a wrong shape
const obj = { 
  status: UserDataProcessingStatusEnum.PENDING, // any but FAILED
  reason: "any string",
  /* other UserDataProcessing fields */
};

// what we intended 
UserDataProcessing.decode(obj)
// --> right({ status: UserDataProcessingStatusEnum.PENDING,  /* other UserDataProcessing fields */ });

// what we have 
UserDataProcessing.decode(obj)
// --> left( /* validation error because reason is not admitted */ );
```

balanza (approved):
LGTM

PR #143, opened 2021-04-23
Upgrade to `@pagopa/openapi-codegen-ts`

balanza:
Replace `italia-utils` package with its new version `@pagopa/openapi-codegen-ts`.

This is needed in order to have generated code to reference `@pagopa/ts-commons` instead of `italia-ts-commons`; that means applications which import `@pagopa/io-functions-commons` will have to use `@pagopa/ts-commons`, too


AleDore (approved):
LGTM

PR #145, opened 2021-04-29
Update `@pagopa/io-ts-commons`

balanza:
Update `@pagopa/io-ts-commons` to latest version.

We also need to update relative `express` types:
* upgrade `@types/express` and its dependencies
* upgrade `helmet` to latest version and removed the dependency to its type package (no more needed).


BurnedMarshal (approved):
LGTM

PR #157, opened 2021-05-13
[#IP-144] Re-model reason field as an optional field

balanza:
I decided to revert what has been made in #142. 

I firmly believed we should model the `reason` field to exist for `FAILED` records only. With such convincement, I led @michaeldisaro into a complex codec definition that would satisfy the need of type-soundness. 

As a matter of fact, it turned out this interface to be at least impractical if not dangerous, even. 

With this PR I'm reverting such changes and define `reason` field as optional for any status. As it makes sense for `FAILED`s only, I also remove such field from creation when needed.

### DIFF
--- a/src/models/user_data_processing.ts
+++ b/src/models/user_data_processing.ts
@@ -1,11 +1,11 @@
 import * as t from "io-ts";
 
-import { withDefault } from "@pagopa/ts-commons/lib/types";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+import { tag } from "@pagopa/ts-commons/lib/types";
 
 import { Container } from "@azure/cosmos";
 import { readableReport } from "@pagopa/ts-commons/lib/reporters";
 import { TaskEither } from "fp-ts/lib/TaskEither";
-import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import {
   CosmosdbModelVersioned,
   RetrievedVersionedModel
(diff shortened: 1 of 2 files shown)


src/models/user_data_processing.ts
```diff
@@ -24,33 +24,32 @@ export const USER_DATA_PROCESSING_COLLECTION_NAME = "user-data-processing";
...
- * Ensure UserDataProcessing IDs are in correct shape
- * The unique identifier of a user data processing request identified by concatenation offiscalCode - choice
- */
-export type UserDataProcessingId = t.TypeOf<typeof UserDataProcessingId>;
-export const UserDataProcessingId = t.brand(
-  t.string,
-  (
-    s: string
-  ): s is t.Branded<string, { readonly IUserDataProcessingIdTag: symbol }> => {
+interface IUserDataProcessingIdTag {
```
michaeldisaro, line 27: Wasn't better to maintain branded type that is a more recent functionality than refinement? Refinement is deprecated in new io-ts versions.

michaeldisaro (commented):
LGTM, but I would have preferred brand instead of refinement.

michaeldisaro (approved):
LGTM

balanza:
An example of issue found during the development of https://github.com/pagopa/io-functions-admin/pull/144
Screenshot 2021-05-13 at 14 49 28

AleDore (approved):
LGTM

PR #158, opened 2021-05-13
[#IP-182] Add Payee to Message with PaymentData

gquadrati:
Add Payee as new field in Message Model as:
- required field while saving new messages with explicit payment data
- optional field while retrieving messages

#### List of Changes

- Add Payee and mode Payment data `definitions.yaml`,
- Refactor Message Model
- tests refactor

#### Motivation and Context

This new field is required by PA's that wants to send a payment message for others PA so payee could be different from service's `organization_fiscal_code`.

unit test

#### Types of changes

#### Checklist:


openapi/__tests__/definitions.test.ts
```diff
@@ -109,3 +121,71 @@ describe("ServicePayload definition", () => {
...
+  const aMessageWithoutContent = {
+    id: "A_MESSAGE_ID",
+    fiscal_code: aFiscalCode,
+    created_at: aDate,
+    sender_service_id: "test"
+  };
+
+  const aCountentWithoutPaymentData = {
+    subject:
+      "A Subject of more than 80 characters. Try to reach this value with stupid words, but it's too hard!!! Very hard!!!!!!",
```
balanza, line 138:
```js
"a".repeat(81)
```
58nucd

gquadrati: ahhahahah

openapi/__tests__/definitions.test.ts
```diff
@@ -176,7 +175,7 @@ describe("NewMessage definition", () => {
     );
   });
 
-  it("should decode NewMessage with content and payment data but without payee", () => {
+  it("should NOT decode NewMessage with content and payment data but without payee", () => {
```
gquadrati, line 178: So are you saying that every message containing a payment data should require a payee? even if the payee is the sender itself?

AleDore: Yep, in case payee is the sender itself we must force explicitly in payee field :)

gquadrati: Ok, I remember we didn't want to make the sender define this field if no needed, because of backward compatibility.

AleDore: This field is optional for sender, but when we save message in our storage we must enforce if it is undefined :D

gquadrati: Why did you say that the payee is optional for sender?  
I mean: `NewMessage` is the model needed by the [message creation API](https://github.com/pagopa/io-functions-services/blob/022914ce3bf322db929ea293cefcff4937052315/openapi/index.yaml#L100).
With this test I understand that NewMessage with content and payment data but without payee should fail the decoding part. So how can a service send a message with content and payment data with optional payee? 
I'm sure I'm missing something, I'm sorry

AleDore: > NewMessage is the model needed by the message creation API.
Nope `NewMessage` is the model used to store document in cosmosDB as well as a `CosmosDBModel`. The message Creation Api refers to a separated [definition](https://github.com/pagopa/io-functions-services/blob/022914ce3bf322db929ea293cefcff4937052315/openapi/index.yaml#L964) :)

gquadrati: Actually the definition you linked is generated by the template. If you watch within the template you can see that NewMessage refers to io-function-commons one [here](https://github.com/pagopa/io-functions-services/blob/022914ce3bf322db929ea293cefcff4937052315/openapi/index.yaml.template#L647)

AleDore: Ok, now I got your point, I was misleading with [this](https://github.com/pagopa/io-functions-commons/blob/a6babf2322731a9cac975b4f2fb2668f9e407d16/src/models/message.ts#L128) type . So, no problem fixed in [287d057](https://github.com/pagopa/io-functions-commons/pull/158/commits/287d057e7825a647b9fcfb53e3781ef95e0d6833) :)

package.json
```diff
@@ -17,8 +17,8 @@
     "test:integration": "yarn build && cd ./__integrations__ && yarn install --frozen-lockfile && yarn start",
     "test:all": "yarn test && yarn test:integration",
     "lint": "eslint . -c .eslintrc.js --ext .ts,.tsx",
-    "generate:all": "npm-run-all generate:*",
-    "generate:definitions": "gen-api-models --api-spec ./openapi/index.yaml --out-dir ./generated/definitions",
+    "generate": "npm-run-all generate:*",
+    "generate:definitions": "rimraf ./generated && gen-api-models --api-spec ./openapi/index.yaml --out-dir ./generated/definitions",
     "generate:templates": "gulp generate:templates",
```
BurnedMarshal, line 22: I've an issue running this command locally
```
$ gulp generate:templates
fs.js:45
} = primordials;
    ^

ReferenceError: primordials is not defined
```
Seems that the current version of gulp is incompatible with the nodejs runtime 12

BurnedMarshal: Adding this resolution into the `package.json` file should solve the problem
```
"resolutions": {
    "graceful-fs": "^4.2.4"
  }
```

balanza (approved):

michaeldisaro (approved):
LGTM.

PR #160, opened 2021-05-14
[IP-144] Don't save reason if status is not FAILED

balanza:
As a consequence of #157 we need to instruct `UserDataProcessingModel` to prevent a `reason` field to leak into an actual document stored in CosmosDB when `status != FAILED`.
A common scenario for this to happen is when, given a failed `UserDataProcessing`, its status is updated by just changing the actual value in the object.

In order to achieve this behaviour, we introduced a `beforeSave` hook to `CosmosdbModelVersioned`. Such hook is optional and, when defined, it processes a document just before being put into the database. Default behaviour is: no modifications.


michaeldisaro:
@balanza I made a similar change using Omit. Take a look here https://github.com/pagopa/io-functions-commons/pull/161/files. Maybe the hook could be better though.

__integrations__/models/user_data_processing.test.ts
```diff
@@ -140,4 +140,104 @@ describe("Models |> UserDataProcessing", () => {
...
+              status: UserDataProcessingStatusEnum.ABORTED
+            })
+          );
+          expect(result.reason).not.toBeDefined();
+        }
+      )
+      .run();
+
+    // update document with FAILED
+    const updates2 = {
```
AleDore, line 197: Should we rename this with a more meaningful name?

balanza: done in https://github.com/pagopa/io-functions-commons/pull/160/commits/2af7d800afeddd5a07e269158cbe1e0e09d024d6

AleDore (approved):
LGTM

PR #161, opened 2021-05-14
[#IP-177] Updated dependencies

michaeldisaro:
#### List of Changes
- Updated dependencies to be compliant with fn-admin ones
- Fixed user_data_processing model to get an optional reason
- Used Omit to skim reason when saving other status than failed

#### Motivation and Context
We want to have a coherent and update project that builds locally when io-functions-admin links it in its dependencies.
We want an optional reason field to show a failure reason.

- yarn build
- yarn lint
- yarn test

#### Types of changes

#### Checklist:


src/models/user_data_processing.ts
```diff
@@ -132,27 +133,32 @@ export class UserDataProcessingModel extends CosmosdbModelVersioned<
...
 
-    //
+    // In order to keep data clean, we skim reason field if status is different than FAILED
+    const validUserDataProcessing = {
+      ...(userDataProcessing.status === UserDataProcessingStatusEnum.FAILED
+        ? userDataProcessing
+        : (userDataProcessing as Omit<
+            UserDataProcessing,
+            typeof USER_DATA_PROCESSING_REASON_FIELD
+          >))
```
balanza, line 153: `Omit` does not what you expected here. It doesn't remove the field from an object, it just tells the type system not to expect such field. So the following it's valid:
```ts
type T = { foo: string; bar: number; };

const t: T = { foo: "hi!", bar: 2 };
const to: Omit<T, "bar"> = t;
console.log("foo -->", to.foo); // "hi!"
console.log("bar -->", to.bar); // 2
//                        ^^^ here compiler complaints about bar not being part of t, but the value is actually present at runtime
```
Try it yourself in the [playground](https://www.typescriptlang.org/play?ssl=1&ssc=1&pln=10&pc=1#code/PTAEBUCcE9UhXAdqALgCwJYGdQFMAeAhgLYAOANrqIYgCaha5UDuahKA5DuQPYDmfXPQzJ0VAEaQezRpFABjHoiw9KAKDUpopKuFABeUAG9QAMx48AXAxSQRfANyhxhSNcTxi43JCcBfBw1FZRRUaz1DE3MrUAAiTABCWIAaZ1drACZQALVgrFCUGIB5YgwUAB5wVNiXSFiAPgNUQLzVXAA6Xj4ACljo0ABaAfqU1B526IBKJxA4xNjcpRVKTv5e2sHh0cL22unQWYy1WdBTs-OLy9OAPVvQNB8qRTIMSjlnikIRFBxCcR54KENogeEDcPZQKRXKEeKZUKlxIDUA9QAA3QjkeBUbDUeQoeAY8iwUiQXCMRChdhwJAoDDEXBqIA).

balanza (changes requested):

balanza:
Besides of the specific implementation, if we want to keep the model consistent we must enforce this rule in every point it could be possibly saved to database.
For versione model we have `create`, `update` and `upsert` methods that do that.

src/models/user_data_processing.ts
michaeldisaro: Ok, you're right. Sometimes I forgot that a compilation error is not actually a missing value.

michaeldisaro: I reverted file so we can merge both our PR.

balanza (approved):

PR #162, opened 2021-05-17
[#IP-177] Added a dependency

michaeldisaro:
#### List of Changes
- Added json-set-map

#### Motivation and Context
We want test to work with locally linked package

- yarn build
- yarn lint
- yarn test

#### Types of changes

#### Checklist:


balanza:
What's the scenario of `json-set-map` to be used?

michaeldisaro:
Jest uses it for tests, and when I run tests with the local project linked in the node_modules folder it complaints about this missing dependency. We can move it in dev-dependencies if we want.

michaeldisaro:
@balanza  if I run yarn test on io-functions-admin with the local version of io-functions-commons I obtain this kind of error:
`
 FAIL  UserDataDeleteOrchestrator/__tests__/handler.test.ts
  ● Test suite failed to run

    Cannot find module 'json-set-map' from 'service.js'

    However, Jest was able to find:
        '../utils/conversions.ts'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'ts', 'tsx', 'node'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

    However, Jest was able to find:
        '../GetProfileActivity/handler.ts'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'ts', 'tsx', 'node'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

    However, Jest was able to find:
        '../handler.ts'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'ts', 'tsx', 'node'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

      3 | import { Option } from "fp-ts/lib/Option";
      4 | 
    > 5 | import { Set } from "json-set-map";
        | ^
      6 | 
      7 | import {
      8 |   FiscalCode,

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:230:17)
      at Object.<anonymous> (../../commons/io-functions-commons/src/models/service.ts:5:1)
`

If I import json-set-map these errors are resolved.

balanza:
> We can move it in dev-dependencies if we want.

This for sure is a better option

---


> if I run yarn test on io-functions-admin

This is actually not a good reason, as you are reversing the dependency direction (you are doing something here because it is needed by one of its dependants).

Anyway, we already have 2 occurrences of `json-set-map` in this very repo, [here](https://github.com/pagopa/io-functions-commons/blob/0818badcdd2fe096fb7875119a3f15bb69a23a0e/src/utils/middlewares/__tests__/azure_user_attributes.test.ts#L21) and [here](https://github.com/pagopa/io-functions-commons/blob/f1d08c5dd7474d9f39ef318348392b6109ef2b92/src/models/service.ts#L5). This is because it's a dependency of a dependency:
```sh
ᐅ yarn why json-set-map
=> Found "json-set-map@1.0.2"
info Reasons this module exists
   - "@pagopa#ts-commons" depends on it
   - Hoisted from "@pagopa#ts-commons#json-set-map"
   - Hoisted from "@pagopa#openapi-codegen-ts#@pagopa#ts-commons#json-set-map"
info Disk size without dependencies: "72KB"
info Disk size with unique dependencies: "72KB"
info Disk size with transitive dependencies: "72KB"
info Number of shared dependencies: 0
```

So it makes sense to explicitly add it as a `devDependency`. Anyway, can we get rid of it? Do we really need it?

michaeldisaro:
I could investigate if we really need it, but this could take time that we need to close ip-153 that is prioritary.
I can suggest to add it as dev dependency, test it and release it so we can close #ip-153, and then, closed prioritary tasks, we can add a new toolchain task to investigate if we need it.

balanza (approved):

PR #167, opened 2021-05-31
[#IP-215] Add AzureAllowBodyPayloadMiddleware

balanza:
Add a middleware that allow a specific payload to be provided only by a specific set of user groups.

It is designed for when it's needed to allow only a subset of users to provide a payload that matches a specific subset of the payload schema.

The use case is: we want restrict the opportunity to send a specific field into a Message content to a limited group of users.

### DIFF
--- a/src/utils/middlewares/__tests__/azure_api_auth.test.ts
+++ b/src/utils/middlewares/__tests__/azure_api_auth.test.ts
@@ -287,6 +287,31 @@ describe("AzureAllowBodyPayloadMiddleware", () => {
     expect(aNonMatchingCodec.decode(aPayload).isLeft()).toBe(true); // test is wrong if it fails
   });
 
+  it("should success if pattern is matched and user do belongs to correct user group", async () => {
+    const headers = {
+      ...someHeaders
+    };
+    const aPayload = { foo: { bar: "baz" } };
+    const mockRequest = {
+      header: jest.fn(lookup(headers)),
+      body: aPayload
+    };
+    const aMatchingCodec = t.interface({
+      foo: t.interface({ bar: t.string })
+    });
+    const allowedGroupSet = new Set([UserGroup.ApiMessageWrite]);
+
+    const middleware = AzureAllowBodyPayloadMiddleware(
+      aMatchingCodec,
+      allowedGroupSet
+    );
+
+    const result = await middleware(mockRequest as any);
+
+    expect(isRight(result)).toBe(true);
+    expect(aMatchingCodec.decode(aPayload).isRight()).toBe(true); // test is wrong if it fails
+  });
+
   it("should success if pattern is not matched", async () => {
     const headers = {
       ...someHeaders
--- a/src/utils/middlewares/__tests__/azure_api_auth.test.ts
+++ b/src/utils/middlewares/__tests__/azure_api_auth.test.ts
@@ -4,8 +4,14 @@
 /* eslint-disable sonarjs/no-identical-functions */
 
 import { isLeft, isRight } from "fp-ts/lib/Either";
+import { failure } from "fp-ts/lib/Validation";
+import * as t from "io-ts";
 
-import { AzureApiAuthMiddleware, UserGroup } from "../azure_api_auth";
+import {
+  AzureAllowBodyPayloadMiddleware,
+  AzureApiAuthMiddleware,
+  UserGroup
+} from "../azure_api_auth";
 
 const anAllowedGroupSet = new Set([UserGroup.ApiMessageWrite]);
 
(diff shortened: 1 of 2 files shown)


src/utils/middlewares/azure_api_auth.ts
```diff
@@ -221,3 +222,60 @@ export const AzureApiAuthMiddleware = (
       )
     );
   });
+
+type AzureAllowBodyPayloadMiddlewareMiddlewareErrorResponses =
```
michaeldisaro, line 226: There is a repetition in this type "...MiddlewareMiddleware...", is it desired?

src/utils/middlewares/__tests__/azure_api_auth.test.ts
```diff
@@ -256,3 +262,107 @@ describe("AzureApiAuthMiddleware", () => {
...
+
+    const result = await middleware(mockRequest as any);
+
+    expect(isLeft(result)).toBe(true);
+    result.fold(
+      _ => expect(_.kind).toBe("IResponseErrorForbiddenNoAuthorizationGroups"),
+      _ => fail("Expecting left")
+    );
+  });
+});
```
michaeldisaro, line 368: Is there any test to check that middleware succeed when pattern is matched and user is in the correct group?

michaeldisaro (commented):
Just a question: Should this middleware be always combined with the authorization middleware to check that the calling user is effectively into the desired group?

src/utils/middlewares/__tests__/azure_api_auth.test.ts
```diff
@@ -256,3 +262,107 @@ describe("AzureApiAuthMiddleware", () => {
...
+      anyGroupSet
+    );
+
+    const result = await middleware(mockRequest as any);
+
+    expect(isRight(result)).toBe(true);
+    expect(aNonMatchingCodec.decode(aPayload).isLeft()).toBe(true); // test is wrong if it fails
+  });
+
+  it("should success if pattern is not matched", async () => {
```
gquadrati, line 290: Repeated test name. It can leads to strange behaviors

```suggestion
  it("should success if pattern is not matched - test nr. 2", async () => {
```

gquadrati (commented):
LGTM except the test name and what @michaeldisaro pointed out

src/utils/middlewares/__tests__/azure_api_auth.test.ts
```diff
@@ -256,3 +262,107 @@ describe("AzureApiAuthMiddleware", () => {
...
+      aNonMatchingCodec,
+      anyGroupSet
+    );
+
+    const result = await middleware(mockRequest as any);
+
+    expect(isRight(result)).toBe(true);
+    expect(aNonMatchingCodec.decode(aPayload).isLeft()).toBe(true); // test is wrong if it fails
+  });
+
```
gquadrati, line 312: Add test that pass if patter is matched and user belongs to an allowed group

```suggestion

  it("should success if pattern is matched and current user belongs to user group", async () => {
    const headers = {
      ...someHeaders
    };
    const aPayload = { foo: { bar: "baz" } };
    const mockRequest = {
      header: jest.fn(lookup(headers)),
      body: aPayload
    };
    const aMatchingCodec = t.interface({
      foo: t.interface({ bar: t.string })
    });
    const anyGroupSet = new Set([UserGroup.ApiMessageWrite]);

    const middleware = AzureAllowBodyPayloadMiddleware(
      aMatchingCodec,
      anyGroupSet
    );

    const result = await middleware(mockRequest as any);

    expect(isRight(result)).toBe(true);
    expect(aMatchingCodec.decode(aPayload).isRight()).toBe(true); // test is wrong if it fails
  });

```

michaeldisaro: Shouldn't be:
```
   const headers = {
      ...someHeaders,
      "x-user-groups": "ApiMessageWrite"
    };
```
?

gquadrati: `someHeaders` already contains `"x-user-groups": "ApiMessageWrite"`

src/utils/middlewares/azure_api_auth.ts
balanza: ooops, thanks!

src/utils/middlewares/__tests__/azure_api_auth.test.ts
balanza: added in https://github.com/pagopa/io-functions-commons/pull/167/commits/2c36b433a2768e4e47d1b7275ace7fb53aa35bcd

balanza: good catch, thanks

src/utils/middlewares/azure_api_auth.ts
balanza: fixed in https://github.com/pagopa/io-functions-commons/pull/167/commits/498e40d9165fefd4dcfdac2cd0f3218a6909023c

src/utils/middlewares/__tests__/azure_api_auth.test.ts
michaeldisaro: Ok, good.

michaeldisaro (approved):
LGTM

src/utils/middlewares/__tests__/azure_api_auth.test.ts
michaeldisaro: I committed it, it was a good suggestion.

balanza:
> Just a question: Should this middleware be always combined with the authorization middleware to check that the calling user is effectively into the desired group?

I actually copied that part :
```ts
NonEmptyString.decode(request.header("x-user-groups"))
```

In the end, it's nothing more than a plain text header

PR #178, opened 2021-06-25
[#IP-262] Add preferencesMode e preferencesVersion into Profile model

BurnedMarshal:


src/models/profile.ts
```diff
@@ -17,20 +18,37 @@ import { IsEmailValidated } from "../../generated/definitions/IsEmailValidated";
...
  */
 export const Profile = t.intersection([
   t.interface({
     // the fiscal code of the citized associated to this profile
-    fiscalCode: FiscalCode
+    fiscalCode: FiscalCode,
+
+    // how the citizen prefers to handle subscriptions to Services
+    // default value is needed to handle citizens that didn't make the choice yet
+    servicePreferencesSettings: withDefault(ServicePreferencesSettings, {
```
BurnedMarshal, line 48: Can we rename this property as `servicesPreferencesSettings` since these settings are for all the services?

BurnedMarshal (commented):
LGTM, with a suggestion.

PR #185, opened 2021-07-01
Bugfix missing LEGACY preferences mode

balanza:
#### List of Changes

* introduce `LEGACY` preferences mode to map Profiles with old schema in [`99b10ab` (#185)](https://github.com/pagopa/io-functions-commons/pull/185/commits/99b10ab078ea862a8b6e3a5c81cecb03525f1d5c)
* test all modes are considered in [`ddeaee8` (#185)](https://github.com/pagopa/io-functions-commons/pull/185/commits/ddeaee8572f735f09f4d808219a5dbbb8865949b)
* add integration tests for negative scenarios in [`6b67655` (#185)](https://github.com/pagopa/io-functions-commons/pull/185/commits/6b67655825cea1b2474dc3c976bf900e86332eff)
* (chore) add PR template in [`5fc1af5` (#185)](https://github.com/pagopa/io-functions-commons/pull/185/commits/5fc1af519b1fcc3f55fc384a95f4044736459575)

#### Motivation and Context

We recognized we need an explicit definition for Citizens whose `Profile` hasn't switched to new preferences mechanism yet.
By doing so, we also realized we no longer need to pass `version` back and forth to clients, as it can be handled as an internal value by the model; hence, we removed such value from api definitions in https://github.com/pagopa/io-functions-commons/pull/185/commits/99b10ab078ea862a8b6e3a5c81cecb03525f1d5c).
We also constrained the set of possible values to be assigned, by enforcing `version=0` to be allowed only with `mode=LEGACY` and viceversa.

On the other hand, to handle `version` increment correctly is still a responsibility to the controller interacting with `Profile` model. 
 
With integration tests

#### Types of changes

#### Checklist:

### DIFF
--- a/__integrations__/models/profile.test.ts
+++ b/__integrations__/models/profile.test.ts
@@ -11,7 +11,6 @@ import { createContext } from "./cosmos_utils";
 import { fromOption } from "fp-ts/lib/Either";
 import { toString } from "fp-ts/lib/function";
 import { ServicesPreferencesModeEnum } from "../../generated/definitions/ServicesPreferencesMode";
-import { NonNegativeInteger } from "@pagopa/ts-commons/lib/numbers";
 
 const aProfile: Profile = Profile.decode({
   acceptedTosVersion: 1,
@@ -153,8 +152,8 @@ describe("Models |> Profile", () => {
     // update document
     const updates = {
       servicePreferencesSettings: {
-        mode: ServicesPreferencesModeEnum.AUTO,
-        version: 1 as NonNegativeInteger
+        mode: ServicesPreferencesModeEnum.AUTO as const,
+        version: 1
       }
     };
     await model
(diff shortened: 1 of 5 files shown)


__integrations__/models/profile.test.ts
```diff
@@ -153,8 +152,8 @@ describe("Models |> Profile", () => {
     // update document
     const updates = {
       servicePreferencesSettings: {
-        mode: ServicesPreferencesModeEnum.AUTO,
-        version: 1 as NonNegativeInteger
+        mode: ServicesPreferencesModeEnum.AUTO as const,
```
balanza, line 155: @BurnedMarshal please note this counter-intuitive behaviour: 

as we didn't define `mode` as `ServicesPreferencesModeEnum`, but we tell it to be either 
```
ServicesPreferencesModeEnum.AUTO | ServicesPreferencesModeEnum.MANUAL
``` 
or 
```
ServicesPreferencesModeEnum.LEGACY
```
 the type system cannot infer the union we defined is complete in regards of `ServicesPreferencesModeEnum`.

That is, as `ServicesPreferencesModeEnum.AUTO` is of type `ServicesPreferencesModeEnum`, it cannot be assigned to `mode` as the latter is defined with a possibly narrower set of values (if not even considered to be a totally different type).

To overcome such, we must tell the specific value we're assigning to be considered `as const`.

src/models/__tests__/profile.test.ts
```diff
@@ -117,6 +118,27 @@ describe("findLastVersionByModelId", () => {
   });
 });
 
+describe("Profile codec", () => {
+  it("should consider all possible ServicesPreferencesMode values", async () => {
+    // This is a safe-guard to programmatically ensure all possible values of ServicesPreferencesModeEnum are considered
+
+    expect.assertions(0);
```
BurnedMarshal, line 125: Do we need this?

BurnedMarshal (approved):
LGTM, apart minor change

__integrations__/models/profile.test.ts
BurnedMarshal: What if we export `as ServicePreferencesSettings` and we use 
```
const updates = {
      servicePreferencesSettings: {
        mode: ServicesPreferencesModeEnum.AUTO,
        version: 1
      } as ServicePreferencesSettings
    };
```

src/models/__tests__/profile.test.ts
balanza: Actually no, I was sure jest would have complained with a tests without assertions. 

Removed in https://github.com/pagopa/io-functions-commons/pull/185/commits/9dc766413db7cb780209af74a1c641286335877a

Thanks

__integrations__/models/profile.test.ts
BurnedMarshal: or `as Profile["servicePreferencesSettings"]` without any new export

PR #186, opened 2021-07-01
Refine `ServicePreferencesSettings` type definition

balanza:
#### List of Changes

* `servicePreferencesSettings.version` is now either `-1` or `NonNegativeInteger`

#### Motivation and Context

This is a refinement which makes the difference between `LEGACY` and non-`LEGACY` service preferences settings more neat. 

Other than being more explicit for readers, there is no much improvement when dealing with `Profile` model alone.

On the other hand, we have benefits when using `Profile` model alongside `ServicePreference`. As `ServicePreference.settingsVersion` is defined as `NonNegativeInteger`, it is now _impossible_ to create a `ServicePreference` from a `Profile` working in `LEGACY` mode, which is a domain concept now constrained into the type system.

**Please note that this PR must be merged before we have profiles with `servicePreferencesSettings` in production**, as values are not compatible with what's before.

Integration tests

#### Types of changes

#### Checklist:


BurnedMarshal (approved):
LGTM

michaeldisaro:
Be carefull that integration tests are broken.

gquadrati (approved):
Looks good to me too

balanza:
> IMHO the only thing that should happen is that non legacy versions start from 0 instead of 1.

What do you mean?

PR #190, opened 2021-08-02
[#IP-320]  Migrate fp-ts version from 1.x to 2.x

fabriziopapi:
Update fp-ts version from 1.x to 2.x

#### List of Changes
Update fp-ts, io-ts, ts-commons and openapigen versions.
Fixed compilation errors due to different fp-ts 2.x version mode.
Fixed unit test compilation errors.

#### Motivation and Context
Reduce technical debt and enable dev to use new fp-ts feature.

#### How Has This Been Tested?

Already existing unit tests.

#### Types of changes

- [ ] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #196, opened 2021-08-20
[#ENTE-79] listLastVersionServices in Service model

BurnedMarshal:
#### List of Changes

Add new method `listLastVersionServices` in `Service` model to get the last version of all services in cosmos

#### Motivation and Context

To generate services webview and services list into the IO website we need do retrieve all the services from cosmos.

Unit test

#### Types of changes

### DIFF
--- a/src/models/service.ts
+++ b/src/models/service.ts
@@ -270,6 +270,10 @@ export class ServiceModel extends CosmosdbModelVersioned<
                 .decode(_.value)
                 .map(services =>
                   some(
+                    // Remap an array of services to a Record with serviceId as key and
+                    // the entire service as value, the value is updated only if the
+                    // processing service is a newer version.
+                    // From that Record we keep only the values (last version of each service).
                     Object.values(
                       services.reduce((prev, curr) => {
                         const isNewer =
--- a/src/models/service.ts
+++ b/src/models/service.ts
@@ -263,30 +263,28 @@ export class ServiceModel extends CosmosdbModelVersioned<
     )
       .map(_ => fromNullable(_.resources))
       .chain(_ =>
-        _.isSome()
-          ? _.value.length > 0
-            ? fromEither(
-                t
-                  .readonlyArray(this.retrievedItemT)
-                  .decode(_.value)
-                  .map(services =>
-                    some(
-                      Object.values(
-                        services.reduce((prev, curr) => {
-                          const isNewer =
-                            !prev[curr.serviceId] ||
-                            curr.version > prev[curr.serviceId].version;
-                          return {
-                            ...prev,
-                            ...(isNewer ? { [curr.serviceId]: curr } : {})
-                          };
-                        }, {} as Record<string, RetrievedService>)
-                      )
+        _.isSome() && _.value.length > 0
+          ? fromEither(
+              t
+                .readonlyArray(this.retrievedItemT)
+                .decode(_.value)
+                .map(services =>
+                  some(
+                    Object.values(
+                      services.reduce((prev, curr) => {
+                        const isNewer =
+                          !prev[curr.serviceId] ||
+                          curr.version > prev[curr.serviceId].version;
+                        return {
+                          ...prev,
+                          ...(isNewer ? { [curr.serviceId]: curr } : {})
+                        };
+                      }, {} as Record<string, RetrievedService>)
                     )
                   )
-                  .mapLeft(CosmosDecodingError)
-              )
-            : fromEither(right(none))
+                )
+                .mapLeft(CosmosDecodingError)
+            )
           : fromEither(right(none))
       );
   }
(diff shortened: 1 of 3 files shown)


src/models/service.ts
```diff
@@ -246,4 +252,42 @@ export class ServiceModel extends CosmosdbModelVersioned<
...
+    Option<ReadonlyArray<RetrievedService>>
+  > {
+    return tryCatch<CosmosErrors, FeedResponse<RetrievedService>>(
+      () => this.container.items.query(`SELECT * FROM c`).fetchAll(),
+      toCosmosErrorResponse
+    )
+      .map(_ => fromNullable(_.resources))
+      .chain(_ =>
+        _.isSome()
+          ? _.value.length > 0
```
balanza, line 267: We can `AND` those conditions and lower the complexity of the procedure

src/models/service.ts
```diff
@@ -246,4 +252,42 @@ export class ServiceModel extends CosmosdbModelVersioned<
...
+                        services.reduce((prev, curr) => {
+                          const isNewer =
+                            !prev[curr.serviceId] ||
+                            curr.version > prev[curr.serviceId].version;
+                          return {
+                            ...prev,
+                            ...(isNewer ? { [curr.serviceId]: curr } : {})
+                          };
+                        }, {} as Record<string, RetrievedService>)
+                      )
```
balanza, line 284: We may need a named function or a comment line to explain what this code wants to achieve, to avoid readers to go though all implementation details to understand.

balanza (changes requested):
I left some minor style comments, but overall I'm worried this api to consume all elements of the collection in memory. By a very rough estimate, given 10k services with 3 versions each, we'll have to process 30k documents in memory.

Shouldn't we find another solution?

src/models/service.ts
BurnedMarshal: Fixed in 9469f4e

src/models/service.ts
BurnedMarshal: Comment added in d25a09c

src/models/service.ts
```diff
@@ -246,4 +252,42 @@ export class ServiceModel extends CosmosdbModelVersioned<
   ): TaskEither<CosmosErrors, Option<RetrievedService>> {
     return super.findLastVersionByModelId([serviceId]);
   }
+
+  public listLastVersionServices(): TaskEither<
+    CosmosErrors,
+    Option<ReadonlyArray<RetrievedService>>
+  > {
+    return tryCatch<CosmosErrors, FeedResponse<RetrievedService>>(
+      () => this.container.items.query(`SELECT * FROM c`).fetchAll(),
```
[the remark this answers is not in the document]
BurnedMarshal: To works properly we need all the documents, so we can't use `getCollectionIterator`. We can use instead `getCollection`. The code will be like this 
```typescript
    return this.getCollection()
      .chain(
        fromPredicate(
          (_): _ is ReadonlyArray<Right<t.Errors, RetrievedService>> =>
            !_.some(isLeft),
          _ => CosmosDecodingError(_.filter(isLeft)[0].value)
        )
      )
      .map(_ => _.map(service => service.value))
      .chain(services =>
        services.length > 0
          ? taskEither.of(
              some(
                Object.values(
                  services.reduce((prev, curr) => {
                    const isNewer =
                      !prev[curr.serviceId] ||
                      curr.version > prev[curr.serviceId].version;
                    return {
                      ...prev,
                      ...(isNewer ? { [curr.serviceId]: curr } : {})
                    };
                  }, {} as Record<string, RetrievedService>)
                )
              )
            )
          : fromEither(right(none))
      );
```

BurnedMarshal:
> I left some minor style comments, but overall I'm worried this api to consume all elements of the collection in memory. By a very rough estimate, given 10k services with 3 versions each, we'll have to process 30k documents in memory.
> 
> Shouldn't we find another solution?

You are right. This query should be used at most one time at day on a dedicated function (`io-functions-services-cache`). We cannot optimize the data selection at the database layer for ComsosDB SQL query limitation.
An idea could be partitioning the elaboration by groups of `serviceId` on multiple query, but I don't know if the effort is justified for this specific task.

src/models/service.ts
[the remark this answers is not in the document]
balanza: Ok, now I see.

Being that, I suggest a refactor using `getCollectionIterator` as it still fetches all the documents, but it does it in chunks.

By implementing an async map/reduce pipeline, and iterating through all chunks, by bet is we can have:
* a reduction of consumed memory, as the handler has to keep, at each iteration, a single chunk and the aggregated partial result
* a procedure less likely to saturate resources as the number of services grows

balanza (approved):
I left some comments about resource usage of this procedure. As @BurnedMarshal pointed out, that might be acceptable due to the usage scenarios; I leave that decision to you.

The PR looks good to me.

BurnedMarshal:
@balanza I do a rebase in ecbf7f2 after the update of `fp-ts` in #190

balanza (approved):

PR #197, opened 2021-08-20
Add `DecodedFeedResponse` type definition

balanza:
#### Motivation and Context

When dealing with a query with more than a result, we need to get access to results' metadata such as `hasMoreResults` and `continuationToken`. Cosmosdb's SDK provide such informations in its [`FeedResponse` response object](https://github.com/Azure/azure-sdk-for-js/blob/e823b26f4600a698eb1c06f48da9b780fb2331a7/sdk/cosmosdb/cosmos/src/request/FeedResponse.ts#L6); so far, we decided to strip metadata off the results, so to have a cleaner data structure (see [here](https://github.com/pagopa/io-functions-commons/blob/0785e81b4aba3e00e37936cacba5305b133cdc08/src/utils/cosmosdb_model.ts#L261)).

We are now going back on this decision, having query results to come with their metadata. `DecodedFeedResponse` is an interface that mimics `FeedResponse` in naming, but adds a couple of features:
* resources type is enforced through validation
* the result set is an array, thus more convenient to handle in map/reduce pipelines (just as it is now)
* consistency between `hasMoreResults` and `continuationToken` in enforced, that is we have no `continuationToken` when `hasMoreResults=false` and viceversa

The introduced interface is not implemented yet.

##### Usage
```ts
public getQueryIterator(
    query: string | SqlQuerySpec,
    options?: FeedOptions
  ): AsyncIterable<DecodedFeedResponse<TR>> {
    const iterator = this.container.items
      .query(query, options)
      .getAsyncIterator();
    return mapAsyncIterable(iterator, feedResponse =>
      feedResponse.resources.map(e => ({
        ...(typeof feedResponse.continuationToken === "string"
          ? {
              continuationToken: feedResponse.continuationToken,
              hasMoreResults: true
            }
          : { continuationToken: undefined, hasMoreResults: false }),
        resource: this.retrievedItemT.decode(e)
      }))
    );
  }
```

#### List of Changes

* Add `DecodedFeedResponse` type definition

The definition hasn't been used yet

#### Types of changes

#### Checklist:

### DIFF
--- a/src/utils/cosmosdb_model.ts
+++ b/src/utils/cosmosdb_model.ts
@@ -142,6 +142,18 @@ const wrapCreate = <TN, TR>(
 };
 
 /**
+ * Define a FeedResponse in which the type fof the resource is strictly enforced through validation
+ */
+export type DecodedFeedResponse<TResource> = ReadonlyArray<
+  {
+    readonly resource: t.Validation<TResource>;
+  } & (
+    | { readonly hasMoreResults: true; readonly continuationToken: string }
+    | { readonly hasMoreResults: false; readonly continuationToken: undefined }
+  )
+>;
+
+/**
  * A persisted data model backed by a CosmosDB client: this base class
  * abstracts the semantics of the CosmosDB API, by providing the shared code
  * for persisting, retrieving and updating a document model.


michaeldisaro (approved):
LGTM

PR #205, opened 2021-09-03
[#IP-362] Added openapi pagination

michaeldisaro:
#### List of Changes

Changed openapi spec for PaginatedResponse to add actual pagination fields.
Added openapi spec for EnrichedMessage and relative Collection to allow for content enrichment.

#### Motivation and Context

These changes are required for messages pagination and enrichment with content.

* yarn build
* yarn test

#### Types of changes

#### Checklist:

### DIFF
--- a/openapi/definitions.yaml
+++ b/openapi/definitions.yaml
@@ -367,7 +367,7 @@ PaginationResponse:
       example: "first_id_of_the_page"
     next:
       type: string
-      description: Contains a value usefull to get the next page
+      description: Contains a value useful to get the next page
       example: "last_id_of_the_page"
 CreatedMessageWithoutContentCollection:
   description: A collection of messages
--- a/openapi/definitions.yaml
+++ b/openapi/definitions.yaml
@@ -247,6 +247,30 @@ CreatedMessageWithoutContent:
     - fiscal_code
     - created_at
     - sender_service_id
+EnrichedMessage:
+  type: object
+  properties:
+    id:
+      type: string
+    fiscal_code:
+      $ref: "#/FiscalCode"
+    time_to_live:
+      $ref: "#/TimeToLiveSeconds"
+    created_at:
+      $ref: "#/Timestamp"
+    sender_service_id:
+      $ref: "#/ServiceId"
+    service_name:
+      type: string
+    organization_name:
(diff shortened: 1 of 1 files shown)


openapi/definitions.yaml
```diff
@@ -333,13 +357,18 @@ PaginationResponse:
       minimum: 1
       description: Number of items returned for each page.
       example: 2
+    hasMoreResults:
+      type: boolean
+      description: This value indicates if there are more results.
+      example: true
+    prev:
+      type: string
+      description: Contains a value usefull to get the previous page
```
AleDore, line 366:
```suggestion
      description: Contains a value useful to get the previous page
```

openapi/definitions.yaml
```diff
@@ -333,13 +357,18 @@ PaginationResponse:
...
+      example: "first_id_of_the_page"
     next:
       type: string
-      description: |-
-        Contains an URL to GET the next results page in the
-        retrieved collection of items.
-      format: uri
-      example: https://example.com/?p=0XXX2
+      description: Contains a value usefull to get the next page
```
AleDore, line 370:
```suggestion
      description: Contains a value useful to get the next page
```

openapi/definitions.yaml
```diff
@@ -247,6 +247,30 @@ CreatedMessageWithoutContent:
...
+    created_at:
+      $ref: "#/Timestamp"
+    sender_service_id:
+      $ref: "#/ServiceId"
+    service_name:
+      type: string
+    organization_name:
+      type: string
+    message_title:
+      type: string
```
AleDore, line 268: I'm thinking about these attributes. Because of retro-compatibility maybe we should keep them optional, but over the last API version these values are always present AFAIK. What do you think about?

michaeldisaro: These attributes will be updated according to rollout. In future I think we'll have also to get rid of PaginatedEnrichedMessageCollection and related models, and maybe also EnrichedMessages, to embrace a new more clear unified model. I would keep them optional until we will get rid of the path param that says to enrich content or not.
When content will be enriched by default, we could make them required.

AleDore (approved):
LGTM

PR #207, opened 2021-09-08
[#IP-362] Message enrichment

michaeldisaro:
#### List of Changes

* modified async iterator mapping to resolve promises
* fixed fillPage for a more generic id type. 

#### Motivation and Context

We need to enrich messages if requested.

* yarn build
* yarn test
* yarn lint

#### Types of changes

#### Checklist:


src/utils/messages.ts
```diff
@@ -143,3 +148,42 @@ export const retrievedMessageToPublic = (
...
+      );
+      const service = O.getOrElse(() => ({} as Service))(x.service);
+      return {
+        ...message,
+        message_title: content.subject,
+        organization_name: service.organizationName,
+        service_name: service.serviceName
+      };
+    })
+  )();
```
AleDore, line 189: Not sure this function should be placed in a common repo, because its implementation could vary based on client needs. @balanza @BurnedMarshal ?

src/utils/async.ts
```diff
@@ -16,6 +16,24 @@ export const mapAsyncIterator = <T, V>(
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    IteratorReturnResult<any> | { readonly done: boolean; readonly value: V }
+  > =>
+    iter.next().then(async (result: IteratorResult<T>) =>
+      // IteratorResult defines that when done=true, then value=undefined
+      // that is, when the iterator is done there is no value to be procesed
+      result.done ? result : { done: false, value: await f(result.value) }
+    )
+});
+
```
gquadrati, line 36: Amazing! Can we replace 

```` typescript
result.done ? result : { done: false, value: await f(result.value) }
````

with 

````typescript
result.done ? result : { ...result, value: await f(result.value) }
````

File: src/utils/messages.ts (hunk 1 above)
Comment (michaeldisaro) [reply src/utils/messages.ts:189]: I've choosen the place where there is also the implementation of the function that creates public messages. But I agree, this implementation should be in function app.

Comment (gunzip) [reply src/utils/messages.ts:189]: I agree with @AleDore

Comment (gquadrati) [inline src/utils/messages.ts:189]: We could get rid of `x` variable

```suggestion
  pipe(
    TE.Do,
    TE.bind("maybeService", () =>
      serviceModel.findLastVersionByModelId([message.sender_service_id])
    ),
    TE.mapLeft(E.toError),
    TE.bind("maybeMessageContent", () =>
      messageModel.getContentFromBlob(blobService, message.id)
    ),
    TE.map({maybeService, maybeMessageContent} => {
      const content = O.getOrElse(() => ({} as MessageContent))(
        maybeMessageContent
      );
      const service = O.getOrElse(() => ({} as Service))(maybeService);
      return {
        ...message,
        message_title: content.subject,
        organization_name: service.organizationName,
        service_name: service.serviceName
      };
    })
  )();
```

gquadrati (commented):
Very useful, thanks! I left a couple of comments

src/utils/messages.ts
michaeldisaro: Applied, but on fn-app repo.

AleDore, line 189:
```suggestion
  pipe(
    TE.Do,
    TE.bind("maybeService", () =>
      serviceModel.findLastVersionByModelId([message.sender_service_id])
    ),
    TE.mapLeft(E.toError),
    TE.bind("maybeMessageContent", () =>
      messageModel.getContentFromBlob(blobService, message.id)
    ),
    TE.map(({ maybeMessageContent, maybeService }) => {
      const subject = pipe(
        maybeMessageContent,
        O.map(content => content.subject),
        O.toUndefined
      );
      const messageService = pipe(
        maybeService,
        O.map(_ => ({
          organizationName: _.organizationName,
          serviceName: _.serviceName
        })),
        O.getOrElseW(() => ({
          organizationName: undefined,
          serviceName: undefined
        }))
      );
      return {
        ...message,
        message_title: subject,
        organization_name: messageService.organizationName,
        service_name: messageService.serviceName
      };
    })
  )();
```

michaeldisaro: Applied on fn-app.

src/utils/async.ts
```diff
@@ -16,6 +16,24 @@ export const mapAsyncIterator = <T, V>(
     )
 });
 
+/**
+ * Async maps over an AsyncIterator
+ */
+export const asyncMapAsyncIterator = <T, V>(
```
balanza, line 22: I argue we don't need it, we can just add an `await` to `mapAsyncIterator`

src/utils/async.ts
michaeldisaro: I tried a few solutions, but there's an union that is complicating things. We can keep this, as seen in other mappings.

src/utils/paging.ts
```diff
@@ -1,23 +1,22 @@
...
   t.interface({
     hasMoreResults: t.boolean,
-    items: t.readonlyArray(t.interface({ id: NonEmptyString }))
+    items: t.readonlyArray(t.interface({ id: t.string }))
   }),
   t.partial({
-    next: NonEmptyString,
-    prev: NonEmptyString
+    next: t.string,
+    prev: t.string
```
balanza, line 13: Why string? is `""` allowed?

balanza (changes requested):

src/utils/async.ts
michaeldisaro: Tried it, but it was not resolving nested promises.

src/utils/messages.ts
```diff
@@ -143,3 +148,42 @@ export const retrievedMessageToPublic = (
...
+  pipe(
+    TE.Do,
+    TE.bind("service", () =>
+      serviceModel.findLastVersionByModelId([message.sender_service_id])
+    ),
+    TE.mapLeft(E.toError),
+    TE.bind("messageContent", () =>
+      messageModel.getContentFromBlob(blobService, message.id)
+    ),
+    TE.map(x => {
```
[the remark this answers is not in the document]
michaeldisaro: Done.

src/utils/paging.ts
michaeldisaro: There are some older models types that have id defined as string. So we had an error passing them as type.

balanza: Shouldn't we decode them?

AleDore (approved):
LGTM

PR #208, opened 2021-09-08
[#IP-362] Message enrichment - add on

balanza:
This PR is related to https://github.com/pagopa/io-functions-commons/pull/207, to discuss about whether we need `asyncMapAsyncIterator` or not


AleDore (approved):
LGTM, I don't see particular issues. Aside of this should we maintain the same signature?

michaeldisaro (approved):
LGTM, nice approach.

PR #211, opened 2021-09-13
[#IP-362] Pagination

michaeldisaro:
#### List of Changes

- Complete refactoring of utilities useful for paging.
- Changed openapi models related to messages pagination.
- New model that is oneOf EnrichedMessage or CreatedMessageWithoutContent
 
PS: Sorry for branch name, it's wrong.

#### Motivation and Context

We want to have a consistent pagination.

- yarn build
- yarn lint
- yarn test

#### Types of changes

#### Checklist:


src/utils/paging.ts
```diff
@@ -50,5 +30,9 @@ export const fillPage = async <T extends { readonly id: string }>(
     O.map(e => e.id),
     O.toUndefined
   );
-  return { hasMoreResults, items, next, prev };
+  return {
+    items,
+    next,
+    prev
+  };
```
AleDore, line 37: Why moving out hasMoreResults from here?

michaeldisaro: Because we don't need it. If `hasMoreResults === false` then `next === undefined`.

michaeldisaro: Explained: if do not have a `next` there are no more results.

src/utils/async.ts
```diff
@@ -174,3 +176,40 @@ export const reduceAsyncIterator = async <T, V>(
...
+    }
+    if (acc.length === pageSize) {
+      break;
+    }
+    // eslint-disable-next-line functional/immutable-data
+    acc.push(await next.value);
+  }
+
+  return { hasMoreResults, results: acc };
+};
```
AleDore, line 215: @balanza @BurnedMarshal ?

openapi/definitions.yaml
```diff
@@ -271,6 +271,14 @@ EnrichedMessage:
...
     - created_at
     - sender_service_id
+    - service_name
+    - organization_name
+    - message_title
+PublicMessage:
+  x-one-of: true
+  allOf:
+    - $ref: "#/CreatedMessageWithoutContent" 
+    - $ref: "#/EnrichedMessage"
```
balanza, line 281: Should it be `PublicMessage`?

src/utils/async.ts
balanza: Nothing to say on the implementation, I have some concerns on naming, though.

I'd avoid the concept of Page; instead I'd go for `asyncIteratorToLimitedArray`, also because this method has nearly the same signature ans body of `asyncIteratorToArray`.

Which makes me think: can we collapse both methods into one? Maybe passing the `size` argument as optional with a default value of `Infinity` would make it work.

michaeldisaro: But there is a concept of "page" because we do not return an array, but a model that contains an array and a "paging property" called `hasMoreResult`. I suggested this name to make the intention clear when we use this.

michaeldisaro: Here we are de facto returning a "page of items" from a iterator. Caller's readability is a lot improved with this name.

openapi/definitions.yaml
michaeldisaro: Well... we pass from a function that is `retrievedToPublicMessage` so I think it could be...

src/utils/async.ts
michaeldisaro: @balanza done on https://github.com/pagopa/io-functions-commons/pull/211/commits/10a4887c90321c1331ea1b0d774ea7d0df593aad

AleDore (approved):
LGTM

PR #221, opened 2021-11-05
[#IC-50] Composed Cosmos Versioned Model

BurnedMarshal:
#### List of Changes

A new model type is created `CosmosdbModelComposedVersioned`.
Refactory of `CosmosdbModelVersioned` to allow override of `createNewVersion` method.

#### Motivation and Context

We need a new model type that handle versioning for documents with a couple externalKey-partitionKey.
This model works correctly with a limited range of externalKey for a specific partition key.

unit test

#### Types of changes

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
+++ b/src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
@@ -159,8 +159,8 @@ describe("upsert", () => {
   });
 });
 
-describe("update", () => {
-  it("should create a new document with explicit version", async () => {
+describe("create", () => {
+  it("should create a new document with version 0", async () => {
     const expectedNextVersion = 0;
     const externalKey = aMyDocument[aModelExternalKeyId];
     const pk = aMyDocument[aModelPartitionField];
@@ -188,8 +188,8 @@ describe("update", () => {
   });
 });
 
-describe("create", () => {
-  it("should create a new document with version 0", async () => {
+describe("update", () => {
+  it("should create a new document with explicit version", async () => {
     const expectedNextVersion = aRetrievedExistingDocument.version + 1;
     const externalKey = aRetrievedExistingDocument[aModelExternalKeyId];
     const pk = aRetrievedExistingDocument[aModelPartitionField];
--- a/src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
+++ b/src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
@@ -246,3 +246,20 @@ describe("findLastVersionByModelId", () => {
   });
 });
 
+describe("typings restrictions", () => {
+  it("should NOT accept modelExternalKeyId equals to partitionKeyId", () => {
+    class MyComposedModel extends CosmosdbModelComposedVersioned<
+      MyDocument,
+      MyDocument,
+      RetrievedMyDocument,
+      typeof aModelExternalKeyId,
+      // @ts-expect-error
+      typeof aModelExternalKeyId
+    > {
+      constructor(c: Container) {
+        super(c, MyDocument, RetrievedMyDocument, aModelExternalKeyId, aModelExternalKeyId);
+      }
+    }
+  });
+});
+
--- a/src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
+++ b/src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
@@ -1,0 +1,248 @@
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+import { CosmosdbModelComposedVersioned, generateComposedVersionedModelId } from "../cosmosdb_model_composed_versioned";
+import * as t from "io-ts";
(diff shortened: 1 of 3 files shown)


balanza (commented):
> This model works correctly with a limited range of externalKey for a specific partition key

Can you explain this a bit more?

---

Couldn't we generalize `CosmosDbVersioneModel` so that it can handle both simply keys (scalar `modelId`) and composed keys (pair `[modelId, partititonKey]`)? By finding a criteria and by overloading `generateVersionedModelId` we can achieve this. I don't actually know if it's a big value anyway, but I feel like we're really doing the same things in both classes.

Also, consider to write some integrations tests, as they may be more eloquent in terms of what this class is intended for.

BurnedMarshal:
> > This model works correctly with a limited range of externalKey for a specific partition key
> 
> Can you explain this a bit more?

If documents with different external keys grows inside **the same** partition key, the queries will be more expensive.
Nevertheless a composed index including external key and partition key is required on all the models that will use this data structure.

> 
> Couldn't we generalize `CosmosDbVersioneModel` so that it can handle both simply keys (scalar `modelId`) and composed keys (pair `[modelId, partititonKey]`)? By finding a criteria and by overloading `generateVersionedModelId` we can achieve this. I don't actually know if it's a big value anyway, but I feel like we're really doing the same things in both classes.
> 

That was my fist idea, but we already have a Model that has different modelId and partitionKey and the signature we needed for `generateVersionedModelId` it's different in our case. The combination of this two reason forced me to implement this abstract model.

> Also, consider to write some integrations tests, as they may be more eloquent in terms of what this class is intended for.

It's my intention do it on the implementation of this class (es. ActivationModel).

src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
```diff
@@ -0,0 +1,265 @@
...
+    if (E.isLeft(result)) {
+      expect(result.left.kind).toBe("COSMOS_ERROR_RESPONSE");
+      if (result.left.kind === "COSMOS_ERROR_RESPONSE") {
+        expect(result.left.error.code).toBe(500);
+      }
+    }
+  });
+});
+
+describe("update", () => {
```
michaeldisaro, line 162: I think this should be "create"

src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
```diff
@@ -0,0 +1,265 @@
...
+        ...aMyDocument,
+        ...someMetadata,
+        id: documentId(externalKey, pk, expectedNextVersion),
+        version: expectedNextVersion
+      });
+    }
+  });
+});
+
+describe("create", () => {
```
michaeldisaro, line 191: I think this should be "update"

src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
```diff
@@ -0,0 +1,265 @@
...
+  it("should create a new document with version 0", async () => {
```
michaeldisaro, line 192: This title seems wrong. From the test I understand that you are "updating" a retrieved document, and this should create a new version as defined in row 193.

michaeldisaro (commented):
I have some doubts on tests.

src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
BurnedMarshal: fixed in 8532fec

src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
BurnedMarshal: fixed in 8532fec

src/utils/__tests__/cosmosdb_model_composed_versioned.test.ts
BurnedMarshal: Yes, the titles of the unit test was not corresponding with the scenario. Fixed in 8532fec
Thanks

michaeldisaro (approved):
LGTM

balanza:
My first thought when I red `CosmosdbModelComposedVersioned` was we were versioning two entities at the same time. Then I looked at the code and I realized that was about having a composite identity for the document. 

I does not occur to me a more eloquent name for the class, nor I think I should nitpick too much on naming. Nevertheless, as this is a base class we may rename it in future with little effort.

My advice is to ad an eloquent doc block describing the purpose of the class.

For the rest: I'm fine with the code, good job!

BurnedMarshal:
> My first thought when I red `CosmosdbModelComposedVersioned` was we were versioning two entities at the same time. Then I looked at the code and I realized that was about having a composite identity for the document.
> 
> I does not occur to me a more eloquent name for the class, nor I think I should nitpick too much on naming. Nevertheless, as this is a base class we may rename it in future with little effort.
> 
> **My advice is to add an eloquent doc block describing the purpose of the class.**
> 
> For the rest: I'm fine with the code, good job!

@balanza I've added some comments in 80be35d

PR #223, opened 2021-11-12
[#IC-89] App can decode unsupported service category

BurnedMarshal:
#### List of Changes

1. `ServiceMetadata` decoder can successfully decode a `Service` with an unknown service category. The `category` will be ignored.
2. `ServicePayload` accept only `CommonServiceMetadata` without category.

#### Motivation and Context

1. The changes to `ServiceMetadata` allow unupdated apps to decode services with unknown category. This is needed to avoid forced mobile app upgrade when a new category is introduced.
2. Only Admin user can create or modify services category. Admin will use `Service` type instead other user will use `ServicePayload` type.

unit test

#### Types of changes

### DIFF
--- a/openapi/definitions.yaml
+++ b/openapi/definitions.yaml
@@ -634,8 +634,15 @@ ServicePayload:
   description: A payload used to create/update a service for a user.
   x-one-of: true
   allOf:
-   - $ref: "#/VisibleServicePayload"    
+   - $ref: "#/VisibleServicePayload"
    - $ref: "#/HiddenServicePayload"
+ExtendedServicePayload:
+  allOf:
+    - $ref: "#/ServicePayload"
+    - type: object
+      properties:
+        service_metadata:
+          $ref: "#/ServiceMetadata"
 CommonServicePayload:
   description: Common properties for a ServicePayload
   type: object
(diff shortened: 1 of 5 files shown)


openapi/definitions.yaml
```diff
@@ -795,6 +806,7 @@ ServiceMetadata:
   allOf:
     - $ref: "#/StandardServiceMetadata"
     - $ref: "#/SpecialServiceMetadata"
+    - $ref: "#/CommonServiceMetadata"
```
fabriziopapi, line 809: A little bit tricky, but i can not think a better way with openapi 2.x. 
Maybe we could simplify this polimorfism when we'll upgrade to 3.x.

fabriziopapi (approved):

BurnedMarshal:
@fabriziopapi A type was modified in 9b2c7d7 to handle SpecialServices in CreateMessage flow of `io-functions-services`.
You can see how is used in [here](https://github.com/pagopa/io-functions-services/pull/166/files#diff-7917c9ac4e63e649516f0d7efc8e234ead0a748931ee127b694423aef83cedf9R406) to handle Special Services INBOX evaluation.

PR #224, opened 2021-11-23
[#IC-66]  Add Legal Data to message model and write permission

fabriziopapi:
Message body needs a new optional block in order to store the info required by "Messaggi a Valore Legale" project.
To allow senders to fill this block, a new permision has been added.

#### List of Changes

Add LegalData definition to openapi
Add ApiMessageWriteWithLegal permission to api auth

#### Motivation and Context

We needs to store the minimum info to retrieve a MVL from a PEC-SERVER.

unit test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/utils/middlewares/azure_api_auth.ts
+++ b/src/utils/middlewares/azure_api_auth.ts
@@ -80,7 +80,10 @@ export enum UserGroup {
   ApiMessageWriteEUCovidCert = "ApiMessageWriteEUCovidCert",
 
   // messages: send messages with payee
-  ApiMessageWriteWithPayee = "ApiMessageWriteWithPayee"
+  ApiMessageWriteWithPayee = "ApiMessageWriteWithPayee",
+
+  // messages: send messages with legal data
+  ApiMessageWriteWithLegal = "ApiMessageWriteWithLegal"
 }
 /* eslint-enable @typescript-eslint/naming-convention */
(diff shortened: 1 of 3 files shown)


src/utils/middlewares/azure_api_auth.ts
```diff
@@ -80,7 +80,10 @@ export enum UserGroup {
   ApiMessageWriteEUCovidCert = "ApiMessageWriteEUCovidCert",
 
   // messages: send messages with payee
-  ApiMessageWriteWithPayee = "ApiMessageWriteWithPayee"
+  ApiMessageWriteWithPayee = "ApiMessageWriteWithPayee",
+
+  // messages: send messages with legal data
+  ApiMessageWriteWithLegal = "ApiMessageWriteWithLegal"
```
balanza, line 86: Do we have constraints on label's lenght? if no, should we go for `ApiMessageWriteWithLegalData`=

balanza (approved):

src/utils/middlewares/azure_api_auth.ts
fabriziopapi: According with Azure doc, max length is 128 chars, so we can add the data prefix.
(https://docs.microsoft.com/en-us/azure/role-based-access-control/custom-roles)

PR #228, opened 2021-12-10
[#IC-66] Add Legal Message Content and Impersonate Service to model

fabriziopapi:
Update model to include the Legal Message data required by the app to display a mvl and the Impersonate Service data required by the fn-services to retrieve the minimum user data neede to impersonate another service. 

#### List of Changes
Add LegalMessageWithContent model
Add ImpersonateService model

#### Motivation and Context
Required by Legal Message project

unit test

#### Types of changes

#### Checklist:


openapi/definitions.yaml
```diff
@@ -367,6 +367,20 @@ MessageResponseWithContent:
...
+      $ref: "#/CreatedMessageWithContent"
+    legal_message:
+      $ref: "#/LegalMessage"
+    notification:
+      $ref: "#/MessageResponseNotificationStatus"
+    status:
+      $ref: "#/MessageStatusValue"
+  required:
+    - message
+    - legal_message
```
AleDore, line 383: Can we use `allOf` with `MessageResponseWithContent`?

fabriziopapi: refactored with allOf. Thank you!

openapi/definitions.yaml
```diff
@@ -1065,4 +1084,98 @@ MessageCategoryPayment:
...
+  properties:
+    sender:
+      type: string
+      minLength: 1
+    recipients:
+      type: string
+      minLength: 1
+    replies:
+      type: string
+      minLength: 1
```
AleDore, line 1160: maybe these properties should be more structured as array of recipient and array of reply

AleDore (approved):
LGTM, we can modify specs as needed with another PR.

openapi/definitions.yaml
```diff
@@ -1065,4 +1084,98 @@ MessageCategoryPayment:
...
+  type: object
+  properties:
+    id:
+      type: string
+    content_type:
+      type: string
+    name:
+      type: string
+    url:
+      type: string
```
gunzip, line 1174: are these all optional?

openapi/definitions.yaml
```diff
@@ -1065,4 +1084,98 @@ MessageCategoryPayment:
...
+ImpersonatedService:
+  type: object
+  properties:
+    service_id:
+      type: string
+    user_groups:
+      type: string
```
gunzip, line 1181: optional?

fabriziopapi: this is optional becouse an Azure subscription could be created without an user owner. We could force them as required returining an error instead of an empty user_groups (It should never occur for a service, so maybe returning an error has a logic).

openapi/definitions.yaml
fabriziopapi: no, only url should be optional. I'll fix them and update the doc for providers.
Thank you!

PR #232, opened 2021-12-30
Refactor Legal Message 'recipients' to an array

fabriziopapi:
Refactor the recipients field of legal message to an array: the backend will be in charge of splitting the data coming from pec-server.

#### List of Changes

Changed 'recipients' legal message field type to array of string

#### Motivation and Context

Remove data manipolation from app to backend.

unit test

#### Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

### DIFF
--- a/openapi/definitions.yaml
+++ b/openapi/definitions.yaml
@@ -1182,8 +1182,10 @@ CertDataHeader:
       type: string
       minLength: 1
     recipients:
-      type: string
-      minLength: 1
+      type: array
+      items:
+        type: string
+        minLength: 1
     replies:
       type: string
       minLength: 1


balanza (commented):
I'm concerned about backward compatibility, but maybe we are sure it isn't used in prod yet.

Besides that, we should consider define a type which is `string | string[]` for the principle on being permissive on what we accept. What do you think?

fabriziopapi:
> I'm concerned about backward compatibility, but maybe we are sure it isn't used in prod yet.
> 
> Besides that, we should consider define a type which is `string | string[]` for the principle on being permissive on what we accept. What do you think?

Your concern is rigth. This is a field used only in output for legal message, so a union type will still be a breaking change.
Maybe we could add a different field, a sort of "recipientsAsArray", to keep the backward compatibility and remove the other when in future version.

PR #239, opened 2022-02-28
[#IC-313]  Update `message-status` model

gquadrati:
#### List of Changes

- Added "isRead" and "isArchived" attributed in MessageStatus model (with `false` as default)
- Updated tests

#### Motivation and Context

Messages pagination and server-side filtering

Unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/src/models/message_status.ts
+++ b/src/models/message_status.ts
@@ -22,11 +22,18 @@ export const MESSAGE_STATUS_MODEL_PK_FIELD = "messageId" as const;
 
 // We cannot intersect with MessageStatus
 // as it is a *strict* interface
-export const MessageStatus = t.interface({
-  messageId: NonEmptyString,
-  status: MessageStatusValue,
-  updatedAt: Timestamp
-});
+export const MessageStatus = t.intersection([
+  t.interface({
+    messageId: NonEmptyString,
+    status: MessageStatusValue,
+    updatedAt: Timestamp
+  }),
+  t.partial({
+    isRead: t.boolean,
+    // eslint-disable-next-line sort-keys
+    isArchived: t.boolean
+  })
+]);
 
 export type MessageStatus = t.TypeOf<typeof MessageStatus>;
 
--- a/src/models/message_status.ts
+++ b/src/models/message_status.ts
@@ -2,7 +2,10 @@ import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import * as t from "io-ts";
 
 import { Container } from "@azure/cosmos";
-import { TaskEither } from "fp-ts/lib/TaskEither";
+import * as TE from "fp-ts/lib/TaskEither";
+import * as O from "fp-ts/lib/Option";
+import { withDefault } from "@pagopa/ts-commons/lib/types";
+import { pipe } from "fp-ts/lib/function";
 import {
   MessageStatusValue,
   MessageStatusValueEnum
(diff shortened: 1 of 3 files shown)


src/models/message_status.ts
```diff
@@ -22,11 +22,18 @@ export const MESSAGE_STATUS_MODEL_PK_FIELD = "messageId" as const;
...
+  t.interface({
+    messageId: NonEmptyString,
+    status: MessageStatusValue,
+    updatedAt: Timestamp
+  }),
+  t.partial({
+    isRead: t.boolean,
+    // eslint-disable-next-line sort-keys
+    isArchived: t.boolean
+  })
```
AleDore, line 35: What if we decide to let these props required with a default ?

gquadrati: You're right, I'll try this way

michaeldisaro: This a good point. By default a message is not read and not archived.

AleDore: We have to check the case that a messageStatus is stored without these props and we perform a retrieve operation that trigger a decode :)

gquadrati: Yes.. I actually don't like very much the idea of storing the two values in the initial steps, when "status" gets changed, but it's more clear when retrieving them

gquadrati: Done in [0b73fe3](https://github.com/pagopa/io-functions-commons/pull/239/commits/0b73fe318cd16416a6772dd52b2623d480e7812e)
Test updated in [670d905](https://github.com/pagopa/io-functions-commons/pull/239/commits/670d9058d57d0f51598b643d78db13f20319c69a)

src/models/message_status.ts
```diff
@@ -64,12 +64,26 @@ export const getMessageStatusUpdater = (
...
   // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 ): MessageStatusUpdater => status =>
-  messageStatusModel.upsert({
-    kind: "INewMessageStatus",
-    messageId,
-    status,
-    updatedAt: new Date()
-  });
+  pipe(
+    messageStatusModel.findLastVersionByModelId([messageId]),
```
AleDore, line 68: Why do we have to retrieve last version before upsert?

gquadrati: Because we cannot upsert "isRead" and "isArchived" without knowing their previous values

AleDore: AFAIK MessageStatusUpdater is called within the ProcessMessage Workflow in order to track if a message is correctly processed, at the same time a citizen cannot trigger a change status (isPending -> true). Btw LGTM to let the updater agnostic related to workflow usage.

gquadrati: Yes, I saw it, but I prefer to be prepared on new statuses and avoid future mistakes

src/models/message_status.ts
```diff
@@ -57,12 +64,26 @@ export const getMessageStatusUpdater = (
...
+    TE.map(
+      O.getOrElse(() => ({
+        messageId,
+        status,
```
AleDore, line 72: Maybe `status` should be overridden in any case if a message-status is already present

gquadrati: ops, my ad. Done in [2f43605](https://github.com/pagopa/io-functions-commons/pull/239/commits/2f43605ca4cbd7ca67cf37324614389ff6a4aa5e)

AleDore:
@gquadrati : it seems integration tests fail

gquadrati:
> @gquadrati : it seems integration tests fail

@AleDore it seems a mailhog test failed. I re-runned integration tests and they worked fine

AleDore (approved):
LGTM

michaeldisaro (approved):

AleDore (approved):
LGTM

PR #241, opened 2022-03-08
[#IC-350] Add fiscalCode to message_status

fabriziopapi:
Adding fiscalCode to message_status collection allow us to query the messages collection starting from a status change.
This query is 8 times more RU cosuming with messageId only.

#### List of Changes

add fiscalCode as optional field to message_status

#### Motivation and Context

We could enrich a message status change coming from the change feed with the data read from the message collection.

#### Types of changes

#### Checklist:


src/models/message_status.ts
```diff
@@ -18,22 +18,29 @@ import {
...
+    status: MessageStatusValue,
+    updatedAt: Timestamp,
+    // eslint-disable-next-line sort-keys
+    isRead: withDefault(t.boolean, false),
+    // eslint-disable-next-line sort-keys
+    isArchived: withDefault(t.boolean, false)
+  }),
+  // fiscalCode is optional due to retro-compatibility
+  t.partial({
+    fiscalCode: FiscalCode
```
balanza, line 41: We broadly use plain fiscal codes on our models, however we agreed on that we would have built our codebase around hashed values instead. Consider doing so.

balanza (approved):
LGTM beside a little consideration on fiscal codes

src/models/message_status.ts
gquadrati: I agree with you, but we need pain fiscal code here because it's used as partition key for querying `messages` collection.

src/models/message_status.ts
```diff
@@ -18,22 +18,29 @@ import {
...
+    messageId: NonEmptyString,
+    status: MessageStatusValue,
+    updatedAt: Timestamp,
+    // eslint-disable-next-line sort-keys
+    isRead: withDefault(t.boolean, false),
+    // eslint-disable-next-line sort-keys
+    isArchived: withDefault(t.boolean, false)
+  }),
+  // fiscalCode is optional due to retro-compatibility
+  t.partial({
```
gquadrati, line 40: Not sure fiscalCode should be optional. 
Can we create a new MessageStatus model having fiscalCode as required and switching to this new model step by step (first step: use it for creation, second step: after batch operation ended, use it for reading too)? Not sure it's feasible @AleDore @michaeldisaro

fabriziopapi: I think we could promote the fiscalCode to mandatory after the data cleaning.

PR #243, opened 2022-03-08
[#IC-351]  Wrap SDK Cosmos as functional (CRUD + Patch)

fabriziopapi:
Allow cosmos SDK to be used as functional programming (fp-ts style). Cover the CRUD and PATCH (partial update) operations for not versioned model.

#### List of Changes

Add CRUD wrapper
Add PATCH wrapper
Add propertiesToArray function to flat a nested object

#### Motivation and Context

We need all the object to perofmr and update (or upsert) operation on cosmos collection. With the wrapper we can modify a single collection field without knowning the entire document. The operations will be callable directly from a pipe or a flow.

unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "typescript": "^4.3.5"
   },
   "dependencies": {
-    "@azure/cosmos": "^3.11.5",
+    "@azure/cosmos": "^3.15.1",
     "@pagopa/ts-commons": "^10.0.1",
     "applicationinsights": "^1.8.10",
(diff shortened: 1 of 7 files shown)


fabriziopapi:
> 

the RoRo/Module is really an intersting solution. I'll try to rework with it.

balanza:
> > 
> 
> the RoRo/Module is really an intersting solution. I'll try to rework with it.

Anyway, solution `a` can be used right now by implementers, without having to change the library

PR #244, opened 2022-03-10
[#IC-354] Update cosmos lib to 3.15.1

fabriziopapi:
Update the azure/cosmos library to provide _patch_ feature used for document partial update

#### List of Changes

Update cosmos version to 3.15.1

#### Motivation and Context

The current version do not implement the _patch_ operation: in order to update a document value we have to retrieve the whole document before. The partial update operation will allow us to increase data cleasing and materialize view update performance

#### How Has This Been Tested?

unit test

#### Types of changes

- [x] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #245, opened 2022-03-10
[#IC-355] Add patch operation to cosmos model

fabriziopapi:
We need the whole object to parform an update operation on cosmos collection. With the _patch_ we can modify a single collection field knowing only the document key. 

#### List of Changes

add 

#### Motivation and Context

We need to perform some puntually field add/update on document to peroferm data cleansing operation and to mantain updated the materlized view. Without patch, we have to query for the document first, increasing time and costs. 

unit test

#### Types of changes

#### Checklist:


fabriziopapi:
depends on https://github.com/pagopa/io-functions-commons/pull/244

fabriziopapi:
depends on https://github.com/pagopa/io-functions-commons/pull/244

src/utils/cosmosdb_model.ts
```diff
@@ -211,6 +212,56 @@ export abstract class CosmosdbModel<
...
+  ): TE.TaskEither<CosmosErrors, TR> {
+    // documentId must be always valued,
+    // meanwhile partitionKey might be undefined
+    const documentId = searchKey[0];
+    const partitionKey = searchKey[1] || documentId;
+    return pipe(
+      partialDocument,
+      R.toArray(),
+      RA.map(entry => ({
+        op: PatchOperationType.add,
```
AleDore, line 236: what if we have to update an existing path? We can safely use `PatchOperationType.add` as op ?

fabriziopapi: Yes, add operation replace existing value:
```
 Add performs one of the following, depending on the target path:
    If the target path specifies an element that does not exist, it is added.
    If the target path specifies an element that already exists, its value is replaced.
    If the target path is a valid array index, a new element will be inserted into the array at the specified index. It shifts existing elements to the right.
    If the index specified is equal to the length of the array, it will append an element to the array. Instead of specifying an index, you can also use the - character. It will also result in the element being appended to the array.
```

src/utils/__tests__/cosmosdb_model.test.ts
```diff
@@ -288,3 +284,66 @@ describe("find", () => {
...
+
+    expect(patchMock).toBeCalledWith(
+      [{ op: "add", path: "/test", value: anotherTest }],
+      undefined
+    );
+    expect(E.isRight(result)).toBeTruthy();
+    if (E.isRight(result)) {
+      expect(result.right.test).toEqual(anotherTest);
+    }
+  });
```
gquadrati, line 312: Since we are mocking the patch result returning a promise with `test: anotherTest` , what is the value of checking `result`?

gquadrati (changes requested):
I suggest to add some integration tests.

src/utils/__tests__/cosmosdb_model.test.ts
fabriziopapi: you are right, removed!

gquadrati (approved):

PR #275, opened 2022-06-01
[#ICC-87] add new ThirdPartyMessageWithContent definition

gquadrati:
#### Types of changes

#### Checklist:

### DIFF
--- a/openapi/definitions.yaml
+++ b/openapi/definitions.yaml
@@ -521,6 +521,15 @@ LegalMessageResponseWithContent:
           $ref: "#/LegalMessageWithContent"
       required:
         - message
+ThirdPartyMessageWithContent:
+  allOf:
+    - $ref: "#/CreatedMessageWithContent"
+    - type: object
+      properties:
+        third_party_message:
+          $ref: "#/ThirdPartyMessage"
+      required:
+        - third_party_message
 MessageResponseWithoutContent:
   type: object
   properties:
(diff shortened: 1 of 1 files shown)


openapi/definitions.yaml
```diff
@@ -1355,6 +1364,34 @@ ImpersonatedService:
...
+        minLength: 1
+      name:
+        type: string
+        minLength: 1
+      url:
+        type: string
+        minLength: 1
+  required:
+    - id
+    - name
```
fabriziopapi, line 1393: I'm not sure if `name`is required.

openapi/definitions.yaml
```diff
@@ -1355,6 +1364,34 @@ ImpersonatedService:
     - service_id
     - user_groups
     - user_email
+ThirdPartyMessage:
+  type: object
+  properties:
```
fabriziopapi, line 1369: We need a generic property that will contain the "flexible" part of the third-party message (the one matching the confiugred json schema).

gquadrati: I thought it could be a not-strinct implementation, so extended properties could be handled without the need of spec definition.
Do you think there is something wrong with it? 
cc @AleDore

gquadrati: In [fb1e012](https://github.com/pagopa/io-functions-commons/pull/275/commits/fb1e0127014af737e69d324886ae90d23bbea584) I've added `additionalProperties: true` to ThirdPartyMessage, so we expect free-form extra data to be present.

This will allow ThirdParty services to include extra properties that will be included in decode even with `t.exact` decoder.

openapi/definitions.yaml
gquadrati: Fixed in [620fdc9](https://github.com/pagopa/io-functions-commons/pull/275/commits/620fdc9165792fdc6fc3644b2b4c3bd13d763879)

fabriziopapi (approved):

PR #284, opened 2022-07-26
[#ICC-167]  Fix get message error 500 when retrieving message without content

fabriziopapi:
The getContentFromBlob method of Message model do not set apart empty blob from not existing blob. A not exisitng blob should not throw and error, becouse a message without content is valid.

#### List of Changes

add new getBlobAsTextWithError to azure utils
change getContentFromBlob method to manage 404 get blob error

#### Motivation and Context

The message model is used to retrieve a message. A Service poll this method via fn-service API to known the status message as part of send message process. This polling must return a message without content if the message is not rejected/refused, but not yet processed.

unit test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/models/__tests__/message.test.ts
+++ b/src/models/__tests__/message.test.ts
@@ -818,6 +818,32 @@ describe("getContentFromBlob", () => {
     getBlobAsTextSpy.mockReset();
   });
 
+  it("should get an error if storage return an error", async () => {
+    const err = {
+      name: "StorageError",
+      message:
+        "Generic Error. \nRequestId:27149c2d-001e-000e-2d04-a12766000000\nTime:2022-07-26T15:27:59.0934919Z",
+      code: "GenericError",
+      statusCode: 500,
+      requestId: "27149c2d-001e-000e-2d04-a12766000000"
+    };
+    const getBlobAsTextSpy = jest
+      .spyOn(azureStorageUtils, "getBlobAsTextWithError")
+      .mockReturnValueOnce(() => TE.left(err));
+
+    const errorOrMaybeMessageContent = await model.getContentFromBlob(
+      blobServiceMock as any,
+      aMessageId
+    )();
+
+    expect(E.isLeft(errorOrMaybeMessageContent)).toBeTruthy();
+    if (E.isLeft(errorOrMaybeMessageContent)) {
+      expect(errorOrMaybeMessageContent.left).toEqual(new Error(err.message));
+    }
+
+    getBlobAsTextSpy.mockReset();
+  });
+
   it("should fail with an error when the retrieved blob is empty", async () => {
     const getBlobAsTextSpy = jest
       .spyOn(azureStorageUtils, "getBlobAsTextWithError")
--- a/src/models/__tests__/message.test.ts
+++ b/src/models/__tests__/message.test.ts
@@ -2,6 +2,7 @@
 
 import * as azureStorage from "azure-storage";
 import * as E from "fp-ts/lib/Either";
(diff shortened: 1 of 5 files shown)


src/utils/azure_storage.ts
```diff
@@ -186,6 +187,35 @@ export const getBlobAsText = (
...
   });
 
+/**
+ * Get a blob content as text (string). In case of error, it will return error details.
+ *
+ * @param blobService     the Azure blob service
+ * @param containerName   the name of the Azure blob storage container
+ * @param blobName        blob file name
+ */
+export const getBlobAsTextWithError = (
```
gquadrati, line 197: How did you test it? Maybe it's better to add an integration test

gquadrati (changes requested):

src/utils/azure_storage.ts
fabriziopapi: added for whole getContentBlob

src/models/__tests__/message.test.ts
```diff
@@ -790,11 +792,36 @@ describe("getContentFromBlob", () => {
...
+        "The specified blob does not exist.\nRequestId:27149c2d-001e-000e-2d04-a12766000000\nTime:2022-07-26T15:27:59.0934919Z",
+      code: "BlobNotFound",
+      statusCode: 404,
+      requestId: "27149c2d-001e-000e-2d04-a12766000000"
+    };
     const getBlobAsTextSpy = jest
-      .spyOn(azureStorageUtils, "getBlobAsText")
-      .mockReturnValueOnce(Promise.resolve(E.left(err)));
+      .spyOn(azureStorageUtils, "getBlobAsTextWithError")
+      .mockReturnValueOnce(() => TE.left(err));
```
gquadrati, line 806: We could add another test with error having a different error code

src/utils/azure_storage.ts
```diff
@@ -186,6 +187,35 @@ export const getBlobAsText = (
...
+        blobService.getBlobToText(
+          containerName,
+          blobName,
+          options,
+          (err, result, _) =>
+            err ? resolve(E.left(err)) : resolve(E.right(fromNullable(result)))
+        );
+      }
+    ),
+    constant
```
gquadrati, line 216: Why not using `TE.tryCatch` here?

src/utils/__tests__/azure_storage.test.ts
```diff
@@ -8,9 +8,12 @@ import { isNone } from "fp-ts/lib/Option";
 import {
   BlobNotFoundCode,
   getBlobAsObject,
-  getBlobAsText
+  getBlobAsText,
+  getBlobAsTextWithError
```
gquadrati, line 12: Why importing it, since it's not used? Is there a non-committed test?

gquadrati (approved):
LGTM, just some small comments

src/utils/azure_storage.ts
fabriziopapi: The tryCatch will route a throwed error in the left side. In this case we had a callback managing the error.
How could we use the tryCatch to manage it?

PS: I tried with TE.taskify, but I could not understand how to make it work.

src/models/__tests__/message.test.ts
fabriziopapi: done. thank you!

src/utils/__tests__/azure_storage.test.ts
fabriziopapi: no, but I'm adding some test. thank you!

PR #289, opened 2022-08-23
[ICC-195] Add rejection_detail to MessageStatus

gquadrati:
#### List of Changes

Add `rejection_detail` to `MessageStatus`when status is `REJECTED`. 

#### Motivation and Context

RFC 9012- We need to distinguish the reason of a rejected message. A message can be rejected because recipient is not a user of IO, or because the service is not allowed to contact that user.

Unit tests

#### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

### DIFF
--- a/src/models/message_status.ts
+++ b/src/models/message_status.ts
@@ -7,9 +7,17 @@ import * as O from "fp-ts/lib/Option";
 import { withDefault } from "@pagopa/ts-commons/lib/types";
 import { pipe } from "fp-ts/lib/function";
 import {
-  MessageStatusValue,
-  MessageStatusValueEnum
-} from "../../generated/definitions/MessageStatusValue";
+  RejectedMessageStatusValue,
+  RejectedMessageStatusValueEnum
+} from "../../generated/definitions/RejectedMessageStatusValue";
+import {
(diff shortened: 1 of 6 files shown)


Garma00 (approved):
lgtm

src/models/message_status.ts
```diff
@@ -57,8 +86,21 @@ export const RetrievedMessageStatus = wrapWithKind(
 
 export type RetrievedMessageStatus = t.TypeOf<typeof RetrievedMessageStatus>;
 
+// --------------------------------------
+// MessageStatusUpdater
+// --------------------------------------
+
 export type MessageStatusUpdater = (
-  status: MessageStatusValueEnum
+  statusUpdate:
```
AleDore, line 94: Why not create a type with the union of these types?

gquadrati: Thanks, done in [222e6fb](https://github.com/pagopa/io-functions-commons/pull/289/commits/222e6fbd487062196b152d9f53f9839a0eb949fd)

fabriziopapi (approved):

PR #293, opened 2022-09-28
[#IOCIT-112] Removed UNSET value from API specs of Reminder Status

michaeldisaro:
#### List of Changes

Removed UNSET value from API specs of Reminder Status and kept it as a model's only property value.

#### Motivation and Context

We don't want the mobile app to deal with the UNSET value. Mobile app will receive an `undefined` when no reminder status has been set while model on cosmos db will track the new property for every profile by setting an `UNSET` default.

* yarn build
* yarn test
* yarn lint

#### Types of changes

#### Checklist:

### DIFF
--- a/openapi/definitions.yaml
+++ b/openapi/definitions.yaml
@@ -774,10 +774,8 @@ IsEmailEnabled:
 ReminderStatus:
   type: string
   x-extensible-enum:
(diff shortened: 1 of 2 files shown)


BurnedMarshal (approved):
LGTM

PR #294, opened 2022-10-03
[ICC-200] Adding ttl to CosmosRecource

Garma00:
#### List of Changes

* Add `ttl` property to `CosmosResource`
* Add `updateTTL` method to `CosmosdbModel`
* Add `updateTTLForAllVersions` method to `CosmosdbModelVersioned`
* Add `findAllVersionsByPartitionKey` method to `CosmosdbModelVersioned`
* Add integration tests
* Refactored integration tests inside `utils/cosmosdb_model.test.ts`

#### Motivation and Context

those changes are needed in order to handle `ttl` with cosmos.

Integration test.
`NB:`
`@zeit/cosmosdb-server` does not support `patch` so I have tested `updateTTL` and `updateTTLForAllVersions` using `io-d-cosmos-free`, then i have commented the test.

#### Types of changes

#### Checklist:

### DIFF
--- a/__integrations__/utils/cosmosdb_model.test.ts
+++ b/__integrations__/utils/cosmosdb_model.test.ts
@@ -11,10 +11,13 @@ import {
   CosmosdbModel,
   CosmosResource
 } from "../../src/utils/cosmosdb_model";
+
+import { CosmosdbModelVersioned, RetrievedVersionedModel } from "../../src/utils/cosmosdb_model_versioned";
 import { createContext } from "../models/cosmos_utils";
 
 const MyDocument = t.intersection([t.interface({
   pk: t.string,
+  id: t.string,
   test: t.string,
 }), t.partial({messageId: t.string})]);
 type MyDocument = t.TypeOf<typeof MyDocument>;
@@ -32,6 +35,21 @@ class MyModel extends CosmosdbModel<
 > {
   constructor(c: Container) {
     super(c, NewMyDocument, RetrievedMyDocument);
+  }
+}
+
+const RetrievedMyVersionedDocument = t.intersection([MyDocument, RetrievedVersionedModel]);
+type RetrievedMyVersionedDocument = t.TypeOf<typeof RetrievedMyVersionedDocument>;
+
+class MyVersionedModel extends CosmosdbModelVersioned<
+  MyDocument,
+  NewMyDocument,
+  RetrievedMyVersionedDocument,
+  "id",
+  "pk"
+>{
+  constructor(c: Container){
+    super(c, NewMyDocument, RetrievedMyVersionedDocument, "id", "pk");
   }
 }
 
@@ -192,6 +210,37 @@ describe("find", () => {
       expect(O.isNone(result.right)).toBeTruthy();
     }
   });
+
+  it("should return 3 documents", async () => {
+    const model = new MyVersionedModel(context.container);
+
+    await model.upsert({
+      id: testId,
+      pk: testPartition,
+      test: "test"
+    })();
+
+    await model.upsert({
+      id: "2testId" as NonEmptyString,
+      pk: testPartition,
+      test: "test"
+    })();
+
+    await model.upsert({
+      id: "3testId" as NonEmptyString,
+      pk: "invalidPartition",
+      test: "test"
+    })();
+
+    const result = await model.findAllVersionsByPartitionKey([testId, testPartition])();
+
+    expect(E.isRight(result)).toBeTruthy();
+    if (E.isRight(result)) {
+      // expect(O.isNone(result.right)).toBeTruthy();
+      expect(result.right).toHaveLength(2)
+    }
+  });
+
 });
 
 afterEach(() => context.dispose())
--- a/__integrations__/utils/cosmosdb_model.test.ts
+++ b/__integrations__/utils/cosmosdb_model.test.ts
@@ -14,6 +14,7 @@ import {
 
 import { CosmosdbModelVersioned, RetrievedVersionedModel } from "../../src/utils/cosmosdb_model_versioned";
 import { createContext } from "../models/cosmos_utils";
+import { NonNegativeNumber } from "@pagopa/ts-commons/lib/numbers";
 
 const MyDocument = t.intersection([t.interface({
   pk: t.string,
@@ -78,7 +79,7 @@ const aDocument = {
 const aDocumentWithTtl = {
   ...aDocument,
   id: anotherTestId,
-  ttl: 300
+  ttl: 300 as NonNegativeNumber
 }
 
 const errorResponse: ErrorResponse = new Error();
--- a/__integrations__/utils/cosmosdb_model.test.ts
+++ b/__integrations__/utils/cosmosdb_model.test.ts
@@ -12,15 +12,21 @@ import {
   CosmosResource
 } from "../../src/utils/cosmosdb_model";
 
-import { CosmosdbModelVersioned, RetrievedVersionedModel } from "../../src/utils/cosmosdb_model_versioned";
+import {
+  CosmosdbModelVersioned,
+  RetrievedVersionedModel
+} from "../../src/utils/cosmosdb_model_versioned";
 import { createContext } from "../models/cosmos_utils";
 import { NonNegativeNumber } from "@pagopa/ts-commons/lib/numbers";
 
-const MyDocument = t.intersection([t.interface({
-  pk: t.string,
-  id: t.string,
-  test: t.string,
-}), t.partial({messageId: t.string})]);
+const MyDocument = t.intersection([
+  t.interface({
+    pk: t.string,
+    id: t.string,
+    test: t.string
+  }),
+  t.partial({ messageId: t.string })
+]);
 type MyDocument = t.TypeOf<typeof MyDocument>;
 
 const NewMyDocument = t.intersection([MyDocument, BaseModel]);
@@ -39,8 +45,13 @@ class MyModel extends CosmosdbModel<
   }
 }
 
-const RetrievedMyVersionedDocument = t.intersection([MyDocument, RetrievedVersionedModel]);
-type RetrievedMyVersionedDocument = t.TypeOf<typeof RetrievedMyVersionedDocument>;
+const RetrievedMyVersionedDocument = t.intersection([
+  MyDocument,
+  RetrievedVersionedModel
+]);
+type RetrievedMyVersionedDocument = t.TypeOf<
+  typeof RetrievedMyVersionedDocument
+>;
 
 class MyVersionedModel extends CosmosdbModelVersioned<
   MyDocument,
@@ -48,8 +59,8 @@ class MyVersionedModel extends CosmosdbModelVersioned<
   RetrievedMyVersionedDocument,
   "id",
   "pk"
->{
-  constructor(c: Container){
+> {
+  constructor(c: Container) {
     super(c, NewMyDocument, RetrievedMyVersionedDocument, "id", "pk");
   }
 }
@@ -80,7 +91,7 @@ const aDocumentWithTtl = {
   ...aDocument,
   id: anotherTestId,
   ttl: 300 as NonNegativeNumber
-}
+};
 
 const errorResponse: ErrorResponse = new Error();
 // eslint-disable-next-line functional/immutable-data
@@ -89,7 +100,6 @@ errorResponse.code = 500;
 jest.setTimeout(60000);
 
 describe("create", () => {
-
   it("should create a document", async () => {
     const context = createContext("id");
     await context.init();
@@ -102,6 +112,7 @@ describe("create", () => {
           ...aDocument
         })
       );
+      expect(result.right).not.toHaveProperty("ttl");
     }
     context.dispose();
   });
@@ -124,7 +135,6 @@ describe("create", () => {
 });
 
 describe("upsert", () => {
-
   it("should create a document", async () => {
     const context = createContext("id");
     await context.init();
@@ -137,6 +147,7 @@ describe("upsert", () => {
           ...aDocument
         })
       );
+      expect(result.right).not.toHaveProperty("ttl");
     }
     context.dispose();
   });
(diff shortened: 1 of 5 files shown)


src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +171,66 @@ export abstract class CosmosdbModelVersioned<
...
+      ),
+      TE.mapLeft(toCosmosErrorResponse)
+    );
+  }
+
+  /**
+   * Given a searchKey returns all the version of a document
+   */
+
+  public findAllVersionByPartitionKey(
```
gunzip, line 208:
```suggestion
  public findAllVersionsByPartitionKey(
```

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +171,66 @@ export abstract class CosmosdbModelVersioned<
...
+            this.getQueryIterator({
+              parameters: [
+                {
+                  name: "@partitionKey",
+                  value: partitionKey
+                }
+              ],
+              query: `SELECT * FROM m WHERE m.${
+                this.partitionKey ? this.partitionKey : this.modelIdKey
+              } = @partitionKey ORDER BY m.version DESC`
```
gunzip, line 225: do we really need to "order by"?

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +171,66 @@ export abstract class CosmosdbModelVersioned<
     );
   }
 
+  /**
+   * Add the given ttl at all the version of the document identified by
+   * the searchKey
+   */
+
+  public updateTTLForAllVersion(
```
gunzip, line 179: I'd rather add a method to patch any field of all versions (general purpose), if you really need it to be here.

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +171,66 @@ export abstract class CosmosdbModelVersioned<
...
+    searchKey: DocumentSearchKey<TR, CosmosDocumentIdKey, PartitionKey>,
+    ttl: NonNegativeNumber
+  ): TE.TaskEither<CosmosErrors, ReadonlyArray<TR>> {
+    return pipe(
+      this.findAllVersionByPartitionKey(searchKey),
+      TE.map(x => pipe(x, RA.rights)),
```
gunzip, line 185:
```suggestion
      TE.map(RA.rights),
```

src/utils/cosmosdb_model.ts
```diff
@@ -264,6 +268,19 @@ export abstract class CosmosdbModel<
     );
   }
 
+  /**
+   * Add or update the ttl of the document identified by the searchKey
+   */
+
+  public updateTTL(
```
gunzip, line 275: why do we want a special patch operation only for the `ttl` field?

Garma00: I thought it could be useful to have a method like that so you can pass a `NonNegativeNumber` directly, but I agree that this could be redundant

src/utils/cosmosdb_model_versioned.ts
Garma00: Actually no, nice catch

src/utils/cosmosdb_model_versioned.ts
Garma00: I do not agree to be honest, i believe that the `ttl` should be the unique property to be updated for all the versions of a document, updating other properties for all the versions would be against the concept of versioning imho ...

gunzip: In this case I wonder if we need this to be here. It looks like we need some data mangling that, I guess, will be run once. Or is this a common need that may arise regularly in the future?

Garma00: This is a common need that we have 👍🏽

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +172,66 @@ export abstract class CosmosdbModelVersioned<
...
+              ([document.id, searchKey[0]] as unknown) as DocumentSearchKey<
+                TR,
+                CosmosDocumentIdKey,
+                PartitionKey
+              >,
+              ttl
+            )
+          )
+        )
+      ),
```
gunzip, line 200: Maybe you can limit nesting using something like:

```suggestion
      TE.map(RA.rights),
      TE.map(
        RA.map(document =>
          super.updateTTL(
            ([document.id, searchKey[0]] as unknown) as DocumentSearchKey<
              TR,
              CosmosDocumentIdKey,
              PartitionKey
            >,
            ttl
          )
        )
      ),
      TE.chain(RA.sequence(TE.ApplicativeSeq)),
```

(I couldn't find a test for this method so I'm unsure : -)

Garma00: As mentioned in the PR description is not that easy to test this method because there is no support for `patch` by@zeit/cosmosdb-server

__integrations__/utils/cosmosdb_model.test.ts
```diff
@@ -133,22 +198,53 @@ describe("find", () => {
...
+    await model.create(aDocument)();
+    const result = await model.find([anotherTestId])();
     expect(E.isRight(result)).toBeTruthy();
     if (E.isRight(result)) {
       expect(O.isNone(result.right)).toBeTruthy();
     }
-    context.dispose();
   });
+
+  it("should return 3 documents", async () => {
```
gquadrati, line 214:
```suggestion
  it("should return all documents belonging to the same partition key", async () => {
```

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +172,66 @@ export abstract class CosmosdbModelVersioned<
...
+    );
+  }
+
+  /**
+   * Given a searchKey returns all the version of a document
+   */
+
+  public findAllVersionsByPartitionKey(
+    searchKey: DocumentSearchKey<TR, ModelIdKey, PartitionKey>
+  ): TE.TaskEither<CosmosErrors, ReadonlyArray<t.Validation<TR>>> {
```
gquadrati, line 211: Since we want to seach by partition key, maybe we can simplify this signature a little by moving from searchKey to partionKey parameter only.
Same suggestion for `updateTTLForAllVersion`

gquadrati (changes requested):

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +172,66 @@ export abstract class CosmosdbModelVersioned<
...
+    ttl: NonNegativeNumber
+  ): TE.TaskEither<CosmosErrors, ReadonlyArray<TR>> {
+    return pipe(
+      this.findAllVersionsByPartitionKey(searchKey),
+      TE.map(RA.rights),
+      TE.chain(documentList =>
+        RA.sequence(TE.ApplicativeSeq)(
+          documentList.map(document =>
+            super.updateTTL(
+              ([document.id, searchKey[0]] as unknown) as DocumentSearchKey<
```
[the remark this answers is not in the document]
Garma00: My bad, anyway this is fixed if we accept a `searchKey` directly as you suggested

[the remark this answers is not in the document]
gquadrati: Actually I was wrong, the best way is what you have already done: use the `searchKey`.
So please fix this code and leave `searchKey` as `updateTTLForAllVersion` parameter.

src/utils/cosmosdb_model_versioned.ts
gquadrati: As said, it's better to leave `seachKey` here. Sorry, my bad :)

src/utils/cosmosdb_model.ts
```diff
@@ -71,6 +72,7 @@ export type CosmosResource = t.TypeOf<typeof CosmosResource>;
 // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
 export const CosmosResource = t.intersection([
   BaseModel,
+  t.partial({ ttl: NonNegativeNumber }),
```
BurnedMarshal, line 75: IMHO this property should be added to the `BaseModel` type because it could be used even on creation (`TN`) and is not a property added by cosmos itself (like the other properties in `TR`).

src/utils/cosmosdb_model.ts
```diff
@@ -264,6 +268,20 @@ export abstract class CosmosdbModel<
...
+  /**
+   * Add or update the ttl of the document identified by the searchKey
+   */
+
+  public updateTTL(
+    searchKey: DocumentSearchKey<TR, CosmosDocumentIdKey, PartitionKey>,
+    ttl: NonNegativeNumber
+  ): TE.TaskEither<CosmosErrors, TR> {
+    // this cast was necessary even making patch accept a Partial<TR>
+    return this.patch(searchKey, { ttl } as {
```
BurnedMarshal, line 280: The patch method should accept a `Partial<TR>` instead of a `Partial<T>`

src/utils/cosmosdb_model.ts
```diff
@@ -79,7 +81,9 @@ export const CosmosResource = t.intersection([
   })
   // this cast is used to keep CosmosResource aligned
   // with @azure/cosmos/Resource type definition
-]) as t.Type<Resource & { readonly id: NonEmptyString }>;
+]) as t.Type<
+  Resource & { readonly id: NonEmptyString; readonly ttl?: NonNegativeNumber }
```
BurnedMarshal, line 85: Building a readonly BaseModel could make easier the future extension of this typing.
```
const ReadonlyBaseModel = t.readonly(BaseModel);
type ReadonlyBaseModel = t.TypeOf<typeof ReadonlyBaseModel>;
```
```suggestion
  Resource & ReadonlyBaseModel
```

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +172,66 @@ export abstract class CosmosdbModelVersioned<
...
+    ttl: NonNegativeNumber
+  ): TE.TaskEither<CosmosErrors, ReadonlyArray<TR>> {
+    const partitionKey = searchKey.length === 1 ? searchKey[0] : searchKey[1];
+    return pipe(
+      this.findAllVersionsByPartitionKey(searchKey),
+      TE.map(RA.rights),
+      TE.map(
+        RA.map(document =>
+          super.updateTTL(
+            ([document.id, partitionKey] as unknown) as DocumentSearchKey<
```
BurnedMarshal, line 191: the cast to `unknown` should be not needed.

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +172,66 @@ export abstract class CosmosdbModelVersioned<
...
+              TR,
+              CosmosDocumentIdKey,
+              PartitionKey
+            >,
+            ttl
+          )
+        )
+      ),
+      TE.chain(RA.sequence(TE.ApplicativeSeq)),
```
BurnedMarshal, line 200: What if some of this operation fail permanently (over the internal cosmos SDK retry)? We could have an inconsistent state when some version are updated and other aren't. Can we consider a batch operation instead? Maybe we can't, seems that transactional batch are not yet supported from the node SDK. https://github.com/Azure/azure-sdk-for-js/issues/13368

src/utils/cosmosdb_model_versioned.ts
BurnedMarshal: We should careful about this function because when it is applied to collections where documents have a lot of versions (aka profile, preferences) it could become very expensive on CPU/memory and RU consumption.

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +172,66 @@ export abstract class CosmosdbModelVersioned<
...
+      TE.mapLeft(toCosmosErrorResponse)
+    );
+  }
+
+  /**
+   * Given a searchKey returns all the version of a document
+   */
+
+  public findAllVersionsByPartitionKey(
```
BurnedMarshal, line 209: ditto

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +172,66 @@ export abstract class CosmosdbModelVersioned<
...
+   * the searchKey
+   */
+
+  public updateTTLForAllVersion(
+    searchKey: DocumentSearchKey<TR, ModelIdKey, PartitionKey>,
+    ttl: NonNegativeNumber
+  ): TE.TaskEither<CosmosErrors, ReadonlyArray<TR>> {
+    const partitionKey = searchKey.length === 1 ? searchKey[0] : searchKey[1];
+    return pipe(
+      this.findAllVersionsByPartitionKey(searchKey),
```
BurnedMarshal, line 186: maybe we don't need to get **ALL** the documents versions, what if we take only the last (cheaper query) and iterate the update on all the documents version from 1 to the last.

src/utils/cosmosdb_model.ts
gquadrati: We tried it, and much more,  to make it compile, but it seems compiler doesn't like it. 
However we should solve it by passing the `ttl` in the BaseModel, as you suggested. We'll try.

src/utils/cosmosdb_model_versioned.ts
gquadrati: Very cleaver! We then can use `generateVersionedModelId` to generate each document id from 0 to last version.

src/utils/cosmosdb_model_versioned.ts
gquadrati: Unfortunately it seems there's no way to do it with node SDK. 
We need to be careful on monitoring, and fix it manually I think.
@AleDore @fabriziopapi

src/utils/cosmosdb_model.ts
AleDore: I agree with @BurnedMarshal

src/utils/cosmosdb_model_versioned.ts
AleDore: I think we can try with latest sdk version if needed. It seems that transactional support has been merged https://github.com/Azure/azure-sdk-for-js/pull/16838

Garma00:
Moved [here](https://github.com/pagopa/io-functions-commons/pull/296)

PR #296, opened 2022-10-14
Icc 200 adding ttl with batch

Garma00:
#### List of Changes

* [Updated `@azure/cosmos` `to 3.17.1`](362f9b4cf595274f0724978243cf509a77b0c7b3);
* [Added ttl with logic to handle it](5d5f5e76f16b2567f54d5b5eac67a786f8330185);
* [Added unit test](47fb4ad6890bb59869798ca45942df8e6ac9b74d)
* [Added integration test](8f600fbf7655141ccd650fb513c0918ae6143917)

#### Motivation and Context

those changes are needed in order to handle ttl with cosmos.

Integration and unit tests.
`NB`:
`@zeit/cosmosdb-server` does not support `patch` so I have tested `updateTTLForAllVersions` using `io-d-cosmos-free`, then i have commented the test.

#### Types of changes

#### Checklist:

### DIFF
--- a/__integrations__/utils/cosmosdb_model.test.ts
+++ b/__integrations__/utils/cosmosdb_model.test.ts
@@ -12,6 +12,10 @@ import {
   CosmosResource
 } from "../../src/utils/cosmosdb_model";
 import { createContext } from "../models/cosmos_utils";
+import {
+  CosmosdbModelVersioned,
+  RetrievedVersionedModel
+} from "../../src/utils/cosmosdb_model_versioned";
 
 const MyDocument = t.interface({
   pk: t.string,
@@ -149,6 +153,61 @@ describe("find", () => {
   });
 });
 
+describe("findAllVersionsByPartitionKey", () => {
+  class MyVersionedModel extends CosmosdbModelVersioned<
+    MyDocument,
+    NewMyDocument,
+    RetrievedMyVersionedDocument,
+    "pk"
+  > {
+    constructor(c: Container) {
+      super(c, NewMyDocument, RetrievedMyVersionedDocument, "pk");
+    }
+  }
+
+  const RetrievedMyVersionedDocument = t.intersection([
+    MyDocument,
+    RetrievedVersionedModel
+  ]);
+
+  type RetrievedMyVersionedDocument = t.TypeOf<
+    typeof RetrievedMyVersionedDocument
+  >;
+
+  it("should return all documents belonging to the same partition key", async () => {
+    const context = createContext("id");
+    await context.init();
+    const model = new MyVersionedModel(context.container);
+
+    await model.upsert({
+      id: testId,
+      pk: testPartition,
+      test: "test"
+    })();
+
+    await model.upsert({
+      id: "2testId" as NonEmptyString,
+      pk: testPartition,
+      test: "test"
+    })();
+
+    await model.upsert({
+      id: "3testId" as NonEmptyString,
+      pk: "invalidPartition",
+      test: "test"
+    })();
+
+    const result = await model.findAllVersionsByPartitionKey([testPartition])();
+
+    expect(E.isRight(result)).toBeTruthy();
+    if (E.isRight(result)) {
+      // expect(O.isNone(result.right)).toBeTruthy();
+      expect(result.right).toHaveLength(2);
+    }
+    context.dispose();
+  });
+});
+
 /**************** @zeit/cosmosdb-server do not support "patch" yet *****************/
 // const anotherTest = "another-test";
 // describe("patch", () => {
@@ -183,28 +242,6 @@ describe("find", () => {
 // });
 
 // describe("updateTTLForAllVersions", () => {
-
-// const RetrievedMyVersionedDocument = t.intersection([
-//   MyDocument,
-//   RetrievedVersionedModel
-// ]);
-
-// type RetrievedMyVersionedDocument = t.TypeOf<
-//   typeof RetrievedMyVersionedDocument
-// >;
-
-// const MESSAGE_ID = "messageId";
-
-// class MyVersionedModel extends CosmosdbModelVersioned<
-//   MyDocument,
-//   NewMyDocument,
-//   RetrievedMyVersionedDocument,
-//   typeof MESSAGE_ID
-// > {
-//   constructor(c: Container) {
-//     super(c, NewMyDocument, RetrievedMyVersionedDocument, MESSAGE_ID);
-//   }
-// }
 
 //   it("should update the ttl for all the versions", async () => {
 //     const context = createContext(MESSAGE_ID);
--- a/__integrations__/utils/cosmosdb_model.test.ts
+++ b/__integrations__/utils/cosmosdb_model.test.ts
@@ -3,7 +3,7 @@ import * as t from "io-ts";
 import * as E from "fp-ts/lib/Either";
 import * as O from "fp-ts/lib/Option";
 
-import { Container, ErrorResponse, ResourceResponse } from "@azure/cosmos";
+import { Container, ErrorResponse } from "@azure/cosmos";
 
 import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import {
@@ -179,5 +179,57 @@ describe("find", () => {
 //         })
 //       );
 //     }
+//   });
+// });
+
+// describe("updateTTLForAllVersions", () => {
+
+// const RetrievedMyVersionedDocument = t.intersection([
+//   MyDocument,
+//   RetrievedVersionedModel
+// ]);
+
+// type RetrievedMyVersionedDocument = t.TypeOf<
+//   typeof RetrievedMyVersionedDocument
+// >;
+
+// const MESSAGE_ID = "messageId";
+
+// class MyVersionedModel extends CosmosdbModelVersioned<
+//   MyDocument,
+//   NewMyDocument,
+//   RetrievedMyVersionedDocument,
+//   typeof MESSAGE_ID
+// > {
+//   constructor(c: Container) {
+//     super(c, NewMyDocument, RetrievedMyVersionedDocument, MESSAGE_ID);
+//   }
+// }
+
+//   it("should update the ttl for all the versions", async () => {
+//     const context = createContext(MESSAGE_ID);
+//     await context.init();
+//     const model = new MyVersionedModel(context.container);
+
+//     const toInsert = { ...aDocument, messageId: testId };
+
+//     // await model.create(toInsert)();
+
+//     await model.upsert(toInsert)();
+//     await model.upsert(toInsert)();
+//     await model.upsert(toInsert)();
+
+//     const result = await model.updateTTLForAllVersions(
+//       [testId],
+//       300 as NonNegativeNumber
+//     )();
+
+//     if (E.isRight(result)) {
+//       console.log("RIGHT: ", result.right);
+//     } else {
+//       console.log("LEFT: ", result.left);
+//     }
+
+//     // context.dispose();
 //   });
 // });
--- a/src/utils/cosmosdb_model.ts
+++ b/src/utils/cosmosdb_model.ts
@@ -7,6 +7,7 @@ import * as TE from "fp-ts/lib/TaskEither";
 import { TaskEither } from "fp-ts/lib/TaskEither";
 import * as t from "io-ts";
 import { PromiseType } from "@pagopa/ts-commons/lib/types";
+import { NonNegativeNumber } from "@pagopa/ts-commons/lib/numbers";
 
 import {
   Container,
@@ -15,6 +16,7 @@ import {
   FeedResponse,
   ItemDefinition,
   ItemResponse,
+  OperationInput,
   PatchOperationType,
   RequestOptions,
   Resource,
@@ -32,9 +34,12 @@ export type CosmosDocumentIdKey = typeof CosmosDocumentIdKey;
 
 // For basic models, the identity field is always the id of cosmos
 export type BaseModel = t.TypeOf<typeof BaseModel>;
-export const BaseModel = t.interface({
-  id: NonEmptyString
-});
+export const BaseModel = t.intersection([
+  t.interface({
+    id: NonEmptyString
+  }),
+  t.partial({ ttl: NonNegativeNumber })
+]);
 
 // The set of keys for a model
 // T might not include base model fields ("id"), but we know they are mandatory in cosmos documents
@@ -210,6 +215,19 @@ export abstract class CosmosdbModel<
       this.retrievedItemT,
       this.container.items.upsert.bind(this.container.items)
     )(newDocument, options);
+  }
+
+  public batch(
+    // eslint-disable-next-line
+    operations: OperationInput[],
+    partitionKey?: NonEmptyString,
+    options?: RequestOptions
+    // eslint-disable-next-line
+  ): TE.TaskEither<CosmosErrors, any> {
+    return TE.tryCatch(
+      () => this.container.items.batch(operations, partitionKey, options),
+      toCosmosErrorResponse
+    );
   }
 
   /**
--- a/src/utils/cosmosdb_model_versioned.ts
+++ b/src/utils/cosmosdb_model_versioned.ts
@@ -3,15 +3,19 @@ import * as t from "io-ts";
 import * as O from "fp-ts/lib/Option";
 import { TaskEither } from "fp-ts/lib/TaskEither";
 import * as TE from "fp-ts/lib/TaskEither";
+import * as RA from "fp-ts/lib/ReadonlyArray";
 
 import {
   INonNegativeIntegerTag,
-  NonNegativeInteger
+  NonNegativeInteger,
+  NonNegativeNumber
 } from "@pagopa/ts-commons/lib/numbers";
 
 import {
   Container,
   ItemDefinition,
+  JSONValue,
+  PatchOperationType,
   RequestOptions,
   SqlQuerySpec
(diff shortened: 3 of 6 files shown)


src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +173,95 @@ export abstract class CosmosdbModelVersioned<
     );
   }
 
+  public updateTTLForAllVersions(
+    searchKey: DocumentSearchKey<TR, ModelIdKey, PartitionKey>,
+    ttl: NonNegativeNumber
+    // eslint-disable-next-line
+  ): TaskEither<CosmosErrors, TaskEither<never, readonly any[]>> {
+    return pipe(
+      this.findAllVersionsByPartitionKey(searchKey),
```
gunzip, line 182: Have you considered the implementation proposed by @BurnedMarshal here?

https://github.com/pagopa/io-functions-commons/pull/294#discussion_r989857265

I guess you don't even need to query for the latest version, just cycle starting from version 0, increment it and stop when the item is not found.

fabriziopapi (approved):

gquadrati (approved):

src/utils/cosmosdb_model_versioned.ts
gunzip: In which case do you get version jumps?

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -166,6 +173,95 @@ export abstract class CosmosdbModelVersioned<
...
+        () =>
+          asyncIterableToArray(
+            this.getQueryIterator({
+              parameters: [
+                {
+                  name: "@partitionKey",
+                  value: partitionKey as JSONValue
+                }
+              ],
+              query: `SELECT * FROM m WHERE m.${
```
gunzip, line 254: If you don't need to get all the fields consider listing them as an alternative to select star, since this will require more network and memory (resulting in a slower query).

src/utils/cosmosdb_model_versioned.ts
fabriziopapi: Manual and scriptef operations (like the data cancellation process) could leave inconsistance data.
We could not enforce this constraints on cosmos, so we have to take care of this use case.

BurnedMarshal: This approach doesn't eliminate the possibility of a "version jump" especially on large collection (the time between obtaining a document and patching may occur a few minutes later). The get and update process are not transactional, so we can have a document deletion while the update process is running.

BurnedMarshal: IMHO the "version jump" it's a data inconsistency problem that should rise an error on the update process.
I'll think about on the reason of versioned model:
1. Keep the documents readonly
2. Have an history of any change (even for legal purpose)
3. other?
So manual and scripted operations that introduce an inconsistent state should be handled properly, and any cross version operation should fail requiring an administrator intervention to fix it.

gunzip: I agree with @BurnedMarshal

gunzip: Moreover you can use the @BurnedMarshal's algorithm (obtain `max(version)` then delete from there to zero) resunting in the same guarantees (even in case of version jumps) but avoiding to query all revisions.

BurnedMarshal: @gunzip for general update operation process data from max to 0 can be used, but for this specific case not, because we are introducing a `ttl` to documents, and a partial update (caused by pagination of max 100 operation in a single transactional batch) can make active an old version.

gquadrati: Since it's just an optimization, and it leads to a less readable error handling, can we postpone it when it became necessary? 
For now we will implement @gunzip' suggestion by fetching ids only. 
This could be reasonable?

fabriziopapi: > So manual and scripted operations that introduce an inconsistent state should be handled properly, and any cross version operation should fail requiring an administrator intervention to fix it.
Currenlty, very few peaple have write-access to database: managing this situation in operations could be nasty.

src/utils/cosmosdb_model_versioned.ts
Garma00: I agree with @gquadrati, this is a good optimization but I would move this to another pr, at the moment is not possible to filter fields because we pass  the query to `getQueryIterator` which calls the `TR.decode` on the results, so we can only pass the result of a `SELECT *`. I think it's cleaner to merge this pr as is and apply this change into another one.

src/utils/cosmosdb_model.ts
```diff
@@ -212,6 +217,19 @@ export abstract class CosmosdbModel<
     )(newDocument, options);
   }
 
+  public batch(
+    // eslint-disable-next-line
+    operations: OperationInput[],
+    partitionKey?: NonEmptyString,
+    options?: RequestOptions
+    // eslint-disable-next-line
+  ): TE.TaskEither<CosmosErrors, any> {
```
BurnedMarshal, line 226: The model return always a validated output. In here we return any response. We are only wrapping the SDK method loosing any model logic.
What is the output? Maybe `TE.TaskEither<CosmosErrors, Response<ReadonlyArray<TR>>>`?

src/utils/cosmosdb_model.ts
```diff
@@ -32,9 +34,12 @@ export type CosmosDocumentIdKey = typeof CosmosDocumentIdKey;
 
 // For basic models, the identity field is always the id of cosmos
 export type BaseModel = t.TypeOf<typeof BaseModel>;
-export const BaseModel = t.interface({
-  id: NonEmptyString
-});
+export const BaseModel = t.intersection([
```
BurnedMarshal, line 37: This change maybe require to modify the `DocumentSearchKey` type. `ttl` shouldn't be selectable like `PartitionKey` or `ModelIdKey`
```typescript
export type DocumentSearchKey<
  T,
  ModelIdKey extends keyof (T & Omit<BaseModel, "ttl">),
  PartitionKey extends keyof (T & Omit<BaseModel, "ttl">) = ModelIdKey
> = (T & Omit<BaseModel, "ttl">)[ModelIdKey] extends string
  ? PartitionKey extends ModelIdKey
    ? [(T & Omit<BaseModel, "ttl">)[ModelIdKey]]
    : PartitionKey extends keyof (T & Omit<BaseModel, "ttl">)
    ? [
        (T & Omit<BaseModel, "ttl">)[ModelIdKey],
        (T & Omit<BaseModel, "ttl">)[PartitionKey]
      ]
    : never
  : never;
```
The same for the Model class
```typescript
CosmosdbModel<
  T,
  TN extends Readonly<T & BaseModel>,
  TR extends Readonly<T & CosmosResource>,
  PartitionKey extends keyof (T & Omit<BaseModel, "ttl">) = CosmosDocumentIdKey
>
```

src/utils/cosmosdb_model_versioned_ttl.ts
```diff
@@ -0,0 +1,153 @@
...
+      TE.chainW(TE.sequenceSeqArray),
+      TE.chainW(responses =>
+        pipe(
+          responses,
+          RA.filter(response => response.code !== 200),
+          TE.fromPredicate(
+            errors => errors.length === 0,
+            _ =>
+              CosmosErrorResponse({
+                message: `Error updating ttl for ${searchKey}`,
```
BurnedMarshal, line 114: Maybe we need more information in here for example what batch fail.

__integrations__/utils/cosmosdb_model.test.ts
```diff
@@ -181,3 +242,33 @@ describe("find", () => {
 //     }
 //   });
 // });
+
+// describe("updateTTLForAllVersions", () => {
```
BurnedMarshal, line 246: this commented code can be removed?

__integrations__/utils/cosmosdb_model.test.ts
```diff
@@ -149,6 +155,61 @@ describe("find", () => {
...
+      id: "3testId" as NonEmptyString,
+      pk: "invalidPartition",
+      test: "test"
+    })();
+
+    const result = await model.findAllVersionsBySearchKey([testPartition])();
+
+    expect(E.isRight(result)).toBeTruthy();
+    if (E.isRight(result)) {
+      // expect(O.isNone(result.right)).toBeTruthy();
```
BurnedMarshal, line 206: ditto

BurnedMarshal (changes requested):

src/utils/cosmosdb_model.ts
Garma00: You are right, fixed it in [c529b4e](https://github.com/pagopa/io-functions-commons/pull/296/commits/c529b4e7961724b6de93d7ed32652c1dd342c3c8)

src/utils/cosmosdb_model.ts
```diff
@@ -212,6 +217,19 @@ export abstract class CosmosdbModel<
     )(newDocument, options);
   }
 
+  public batch(
+    // eslint-disable-next-line
+    operations: OperationInput[],
```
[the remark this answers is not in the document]
Garma00: Nice point, we have moved this method inside the `CosmosdbModelVersionedTTL` as `private`, it is only used by the `updateTTLForAllVersions` so it just accept `ReadonlyArray<PatchOperationInput>` at the moment, it can be extended in future.

src/utils/cosmosdb_model.ts
Garma00: We moved `ttl` out of `BaseModel` so we don't need huge changes.

src/utils/cosmosdb_model_versioned_ttl.ts
AleDore: Can we add `errors` info here?

gquadrati: Do we need to see all the status codes of the failed chunks? Is it useful?

BurnedMarshal: Chunks are processed sequentially, so only one can fail. The information we need is what chunk fail, to understand how must be handled the DB inconsistency (previous chunks are applied and future aren't)

src/utils/cosmosdb_model_versioned_ttl.ts
```diff
@@ -0,0 +1,207 @@
...
+            this.getQueryIterator({
+              parameters: [
+                {
+                  name: "@partitionKey",
+                  value: partitionKey as JSONValue
+                }
+              ],
+              query: `SELECT * FROM m WHERE m.${
+                this.partitionKey ? this.partitionKey : this.modelIdKey
+              } = @partitionKey`
```
BurnedMarshal, line 166: To ensure that documents are processed on the right order, we need to sort by version
```suggestion
              query: `SELECT * FROM m WHERE m.${
                this.partitionKey ? this.partitionKey : this.modelIdKey
              } = @partitionKey ORDER BY m.version`
```

src/utils/cosmosdb_model_versioned_ttl.ts
BurnedMarshal: Add a comment in [here](https://github.com/pagopa/io-functions-commons/pull/296#discussion_r1000303300) that have an impact on this scenario.

BurnedMarshal: Sorry I'm loosing something. Failed batch operations with `status !== 200` don't interrupt the execution chain. IMHO a failed batch must interrupt the execution with an error.

src/utils/cosmosdb_model.ts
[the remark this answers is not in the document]
gquadrati: We can't narrow it down any more,` ReadonlyArray<PatchOperationInput>` it's the minimum.

src/utils/cosmosdb_model_versioned_ttl.ts
gquadrati: You'r right, even if we get a fail in one chunk (status != 200) we continue to process next chunks (if any).
The target, in my opinion, is to process as many chunks as we can. But we can move the check within the task executed sequentially, to be sure to stop as soon as we get an error.

__integrations__/utils/cosmosdb_model.test.ts
gquadrati: Removed in [4f31c4d](https://github.com/pagopa/io-functions-commons/pull/296/commits/4f31c4da58599e82e5ba5769adbbe0dd0a1affca)

__integrations__/utils/cosmosdb_model.test.ts
gquadrati: Removed in [4f31c4d](https://github.com/pagopa/io-functions-commons/pull/296/commits/4f31c4da58599e82e5ba5769adbbe0dd0a1affca)

src/utils/cosmosdb_model_versioned_ttl.ts
BurnedMarshal: > process as many chunks as we can

I disagree on this, especially for the TTL logic. If a chunk that include the **last version** of the document succeded and any other not, the unrepaired inconsistency (before that the ttl logic take effect) will make active an old version of the document.

src/utils/cosmosdb_model_versioned_ttl.ts
gquadrati: In order not to weigh down the query I have preferred to order version at runtime (we know performances are even better).
Done in [c7d792a](https://github.com/pagopa/io-functions-commons/pull/296/commits/c7d792ad91044f66c0e186aad443e2721426a5b8)

src/utils/cosmosdb_model_versioned_ttl.ts
gquadrati: Done in [ee1803e](https://github.com/pagopa/io-functions-commons/pull/296/commits/ee1803e704acff65bd0231a7b5a8d0dfe7088159)

src/utils/cosmosdb_model_versioned_ttl.ts
michaeldisaro: I was just checking to inform about possibile order by issue for not indexed properties. Remember the  infamous `order by id` for messages? 🤔

src/utils/cosmosdb_model_versioned_ttl.ts
```diff
@@ -0,0 +1,234 @@
...
+          pipe(
+            chunk,
+            RA.toArray,
+            operations => this.batch(operations, partitionKey),
+            TE.chainW(
+              TE.fromPredicate(
+                response => response.code === 200,
+                _ => {
+                  const firstChunkId = chunk[0]?.id ?? -1;
+                  const lastChunkId = chunk[chunk.length - 1]?.id ?? -1;
```
BurnedMarshal, line 126: A empty chunk don't make sense in here. We are executing a batch without any operation? Skip the batch operation if chunk is an empty array instead.
```suggestion
                  const firstChunkId = chunk[0].id;
                  const lastChunkId = chunk[chunk.length - 1].id;
```

src/utils/cosmosdb_model_versioned_ttl.ts
```diff
@@ -0,0 +1,234 @@
...
+                name: `Error updating ttl`
+              })
+          ),
+          TE.chainW(() =>
+            pipe(
+              responses,
+              RA.reduce(0, (v, r) => v + (r.result?.length ?? 0)),
+              TE.right
+            )
+          )
```
BurnedMarshal, line 159: This as been checked before.
```suggestion
        pipe(
          responses,
          RA.reduce(0, (v, r) => v + (r.result?.length ?? 0)),
          TE.right
        )
```

BurnedMarshal (approved):
LGTM. Only the last 2 improvements. Thanks to all the contributors.

src/utils/cosmosdb_model_versioned_ttl.ts
gquadrati: Fine, but we need to make sure that `findAllVersionBySearchKey` did not return an empty array.

src/utils/cosmosdb_model_versioned_ttl.ts
gquadrati: You're right, I just put a error for that should never be raised ("WARNING: This error should never happen!") just for check.
Do you think it's safe to remove it?

src/utils/cosmosdb_model_versioned_ttl.ts
gquadrati: That's why we didn't sort the query, but the results at runtime :D 
But please, remember that we are using ORER BY in `findLastVersionByModelId`

src/utils/__tests__/cosmosdb_model_versioned_ttl.test.ts
```diff
@@ -0,0 +1,239 @@
...
+          result: operations.map(_ => ({ statusCode: 404 })),
+          code: 207
+        };
+      }
+    );
+
+    const model = new MyModel(container);
+    const result = await model.updateTTLForAllVersions(
+      [aModelIdValue],
+      42 as NonNegativeInteger
```
michaeldisaro, line 181: As "theAnswerToLifeUniverseAndEverything".

michaeldisaro (approved):

src/utils/__tests__/cosmosdb_model_versioned_ttl.test.ts
gquadrati: ;)

PR #315, opened 2023-01-26
[#ICC-314] Migrate node 18

Garma00:
#### List of Changes

*  Update node to 18.13;
* Update @types/node to 18;

#### Motivation and Context

End of life Node 14

#### Types of changes

#### Checklist:


fabriziopapi (approved):

fabriziopapi:
LGTM, just a minor sylish comment

src/utils/azure_storage.ts
```diff
@@ -254,7 +254,7 @@ export const getBlobAsObject = async <A, O, I>(
           )
         );
       } catch (e) {
-        return E.left(e);
+        return E.left(new Error(`${e}`));
```
BurnedMarshal, line 257: This is not equivalent because we are generating a new error with message equal to the stringified catched error.
If the error is always an `Error` object we can use the `as` otherwise this must be handled.

BurnedMarshal (changes requested):

src/utils/azure_storage.ts
Garma00: You are right, according to the `JSON.parse()` documentation this method only throws with a [SyntaxError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) which is a subclass of `Error`, as you said we can safely use a cast here IMHO.

Garma00: @BurnedMarshal done in [17eb856](https://github.com/pagopa/io-functions-commons/pull/315/commits/17eb85686a7abeec27e9a8f1b65348673071f2d6)

__integrations__/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "devDependencies": {
     "@types/jest": "^25.2.1",
-    "@types/node": "^13.11.0",
+    "@types/node": "^18.11.18",
```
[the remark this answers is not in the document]
Garma00: Done in [0ec804a](https://github.com/pagopa/io-functions-commons/pull/315/commits/0ec804a63ddb8f9ae874b94ae538396710b15039)

BurnedMarshal (approved):
LGTM

PR #325, opened 2023-03-02
[IOAPE-195] Add authorized cidrs in azure user attribute manage

giuseppedipinto:
Add subscriptionCIDRs model on AzureUserAttributeManage middleware.

#### List of Changes

- Add SubscriptionCIDRs Model
- Updated AzureUserAttributeManage middleware

#### Motivation and Context

We need to limit the use of Manage Subscription key to an authorized IP List *(in CIDRs format)*.

Unit Tests, Integration Tests

#### Types of changes

#### Checklist:


src/models/authorized_cidrs.ts
```diff
@@ -0,0 +1,46 @@
...
+  CosmosResource
+]);
+export type RetrievedAuthorizedCIDRs = t.TypeOf<
+  typeof RetrievedAuthorizedCIDRs
+>;
+
+/**
+ * A model for handling Authorized CIDRs
+ */
+export class AuthorizedCIDRsModel extends CosmosdbModel<
```
balanza, line 32: Is this the case for a versioned model? We may want to audit changes in those values.

src/models/authorized_cidrs.ts
```diff
@@ -0,0 +1,46 @@
...
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+import { readonlySetType } from "@pagopa/ts-commons/lib/types";
+import { CosmosdbModel, CosmosResource } from "../utils/cosmosdb_model";
+import { CIDR } from "../../generated/definitions/CIDR";
+
+export const AUTHORIZED_CIDRS_COLLECTION_NAME = "subscription-cidrs";
+const AUTHORIZED_CIDRS_MODEL_PK_FIELD = "id";
+
+/** List of Authorized IP (in CIDRs format) to allow *subscription keys* use */
+export const AuthorizedCIDRs = t.interface({
```
balanza, line 12: Why do we name the model so different than the collection? Will it make our life easier (adding a useful abstraction for developers) or harder (we have to remember that collection and model are named different)

src/utils/middlewares/azure_user_attributes_manage.ts
```diff
@@ -86,17 +101,71 @@ export const AzureUserAttributesManageMiddleware = (): IRequestMiddleware<
...
-      kind: "IAzureUserAttributesManage"
-    };
-
-    return E.right<
-      IResponse<
-        "IResponseErrorForbiddenNotAuthorized" | "IResponseErrorInternal"
-      >,
-      IAzureUserAttributesManage
-    >(authInfo);
+    // TODO: The condition MUST be removed after io-function-services update
```
balanza, line 104: Please explain `MUST`. Do you see possible inconsitencies?

src/utils/middlewares/azure_user_attributes_manage.ts
```diff
@@ -86,17 +101,71 @@ export const AzureUserAttributesManageMiddleware = (): IRequestMiddleware<
...
+
+      return E.right<
+        IResponse<
+          "IResponseErrorForbiddenNotAuthorized" | "IResponseErrorInternal"
+        >,
+        IAzureUserAttributesManage
+      >(authInfo);
+    } else {
+      const authInfo: IAzureUserAttributesManage = {
+        authorizedCIDRs: new Set((["0.0.0.0/0"] as unknown) as ReadonlyArray<
```
balanza, line 155: I don't think we should make this assumption. We are converting one misconfiguration (to forget to pass a model instance) into a security flaw (we are allowing every IP).

One question we can make to ourselves is: **is it possible, in our domain, that we have a Manage Key and we cannot look up for its cirds?** If the answer is `no`, we should consider two options:

* to make such case irrepresentable 
* to fail the operation immediately

The first to be preferred.

balanza (changes requested):
I have some hard concerns about the implementation and some little comments on naming. 

Also, as we are introducing a db model, consider write an integration test on it.

src/models/authorized_cidrs.ts
giuseppedipinto: Right, we will refactor AuthorizedCIDRs as SubscriptionCIDRs

src/utils/middlewares/azure_user_attributes_manage.ts
giuseppedipinto: The "if" condition on authorizedCIDRsModel (and related else) is temporary because we need to deploy this development in order to update `io-functions-services` impacted functions (UpdateService, UpdateServiceLogo, ...).
Because in those functions we have to update AzureUserAttributeManage with new authorizedCIDRs model.
So the temporary behaviour is useful to have consistency in MANAGE flow during the time we need to update `io-functions-services`.

balanza: I don't understand the need. I only see two possible state for the target function app (`io-functions-services` for the example):
1. it has the **current** version of `io-functions-commons` installed, and **it does not** lookup for cidrs
2. it has the **new** version of `io-functions-commons` installed, and **it does** lookup for cidrs

Do you think we need a third, intermediate state?

src/utils/middlewares/azure_user_attributes_manage.ts
```diff
@@ -86,7 +101,43 @@ export const AzureUserAttributesManageMiddleware = (): IRequestMiddleware<
...
+      );
+    }
+
+    const maybeAuthorizedCIDRs = errorOrMaybeAuthorizedCIDRs.right;
+
+    if (isNone(maybeAuthorizedCIDRs)) {
+      winston.error(
+        `IAzureUserAttributesManage|CIDRs not found|${subscriptionId}`
+      );
+      return E.left<
```
balanza, line 131: We are stating that a manage key MUST have IP protection. Is this correct?

src/utils/middlewares/azure_user_attributes_manage.ts
```diff
@@ -86,7 +101,43 @@ export const AzureUserAttributesManageMiddleware = (): IRequestMiddleware<
...
+        IResponse<"IResponseErrorForbiddenNotAuthorized">,
+        IAzureUserAttributesManage
+      >(ResponseErrorForbiddenNotAuthorized);
+    }
+
     const authInfo: IAzureUserAttributesManage = {
+      authorizedCIDRs: maybeAuthorizedCIDRs.value.cidrs
+        ? maybeAuthorizedCIDRs.value.cidrs
+        : new Set((["0.0.0.0/0"] as unknown) as ReadonlyArray<CIDR>),
```
balanza, line 140: How is `maybeAuthorizedCIDRs.value.cidrs` undefined? What's the meaning of this field being undefined?

src/utils/middlewares/azure_user_attributes_manage.ts
giuseppedipinto: Yes, it's our security goal for this development.
In addition, it should be noted that when we will create a Subscription _(MANAGE)_, we will also create a starting value for subscription related cidrs _(empty array as for services auth_cidrs)._

src/utils/middlewares/azure_user_attributes_manage.ts
giuseppedipinto: This is a typo of the previous commit, for what has been said, it is not possible that the cidrs are undefined, if you agree we will update this line of code removing undefined possibility.

balanza: The type system should have warned you (line 140 should never be reachable. Hasn't it?

src/utils/middlewares/azure_user_attributes_manage.ts
balanza: Uhm, I am not sure we need to enforce such a rule. We do not, so far, with actual API keys.

I think we are solving the problem:

> Users that currently **decided to use** IP protection, want to do the same with manage keys

which is slightly different than:

> Users that will use manage keys will be **required to use** IP protection

----

Moving a bit forward: should we enforce users to provide IP protection? Is it good for the product? We can discuss it (cc @iwoak)

balanza (approved):

PR #331, opened 2023-04-05
[#IOPID-127] Porting `content-digest` crypto fn from `io-functions-lollipop`

BurnedMarshal:
#### List of Changes

Copy crypto functions from `io-functions-lollipop` and `io-backend`.

#### Motivation and Context

Avoid code duplication across repos.

unit test

#### Types of changes

- [X] Chore (nothing changes by a user perspective)

### DIFF
--- a/src/utils/crypto.ts
+++ b/src/utils/crypto.ts
@@ -71,7 +71,7 @@ export const validateDigestHeader = (
   }
 
   // Validate
-  const contentDigestPattern = new RegExp("(.+)=:(.+):");
+  const contentDigestPattern = new RegExp("(sha-256|sha-512)=:(.+):");
   const contentDigestParts = contentDigestPattern.exec(contentDigestHeader);
   if (!contentDigestParts || contentDigestParts.length === 0) {
     throw new Error("Content-digest header invalid");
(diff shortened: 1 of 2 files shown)


src/utils/crypto.ts
```diff
@@ -0,0 +1,96 @@
...
+export const validateDigestHeader = (
+  contentDigestHeader: string,
+  body: Buffer | string
+): void => {
+  if (!contentDigestHeader) {
+    throw new Error("Content-Digest header missing");
+  }
+
+  // Validate
+  const contentDigestPattern = new RegExp("(.+)=:(.+):");
```
BurnedMarshal, line 74: We can reduce the regex vulnerability with this
```suggestion
  const contentDigestPattern = new RegExp("(sha-256|sha-512)=:(.+):");
```

src/utils/crypto.ts
```diff
@@ -0,0 +1,96 @@
+import * as crypto from "crypto";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+
+/**
+ * Returns hashed input based on sha256 algo
+ */
+export const toHash = (s: string): NonEmptyString =>
```
gquadrati, line 7: Maybe `toSha256` is more accurate? @michaeldisaro

gquadrati (approved):
Just a small comment. Agree with regex refactor.

src/utils/crypto.ts
BurnedMarshal: We changed the fn name to `sha256` to keep the same name used into the `io-backend`.

PR #370, opened 2024-03-15
[#DEVEX-60] Update fp-ts and io-ts peerDeps

BurnedMarshal:
#### List of Changes

Upgrade `fp-ts` and `io-ts` to the lastest available version.
Updated other libs in dependencies.

#### Motivation and Context

To complete an integration into a monorepo, is required that `@pagopa/ts-commons` and `@pagopa/io-functions-commons` use the same version of the fp-ts ecosystem libraries. Declaring the two libraries as `peerDependencies` allow the main project that use `io-functions-commons` to command how lib version should be used.

not tested

#### Types of changes

- [X] Chore (nothing changes by a user perspective)


package.json
```diff
@@ -56,14 +56,14 @@
   "dependencies": {
     "@azure/cosmos": "^3.17.1",
     "@azure/data-tables": "^13.2.2",
-    "@pagopa/ts-commons": "^11.0.0",
+    "@pagopa/ts-commons": "^12.5.1",
     "applicationinsights": "^1.8.10",
     "azure-storage": "^2.10.5",
     "cidr-matcher": "^2.1.1",
-    "fp-ts": "^2.10.5",
+    "fp-ts": "2.16.2",
```
balanza, line 63: **question:** ‏We're referencing a specific version rather than a pattern. Is that deliberate or a typo?

balanza (approved):
I left a question.

Anyway, I'm ok with changes as long as they are intended.

package.json
BurnedMarshal: It's deliberate. The last upgrade of this library and the combination with the `io-ts` are incompatible. To avoid compilation error you must have the right combo between the two libs. I've choose to lock the libs version.

balanza: Ok, great. Just for future reference, do you have any issue/article to link, if any?

lucacavallaro (approved):

package.json
BurnedMarshal: The commit [4aa6c4e](https://github.com/pagopa/io-functions-commons/pull/370/commits/4aa6c4ea04b958dacb5c57991cf312dc8eeb4fe8) use `fp-ts` and `io-ts` as peer dependencies. This allow to resolve conflicts at the project level. Now the libs version are not locked anymore.

PR #371, opened 2024-03-27
BREAKING CHANGE: [DEVEX-106] Update Cosmos SDK

lucacavallaro:
#### List of Changes

1. Update `@azure/cosmos` to `4.x` (its latest major)
2. Add `.yarnrc.yml` to fix the yarn version. Without this change, in development environments configured with newer yarn version - yarn tries to update (and auto-migrate) the project to `4.x`.

⚠️ the projects that make use of the base model defined into `@pagopa/io-functions-commons` must be update its `@azure/cosmos` version too, as the previous major is not compatible

Unit test & integration test 🟢 

#### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:


src/utils/cosmosdb_model_versioned.ts
```diff
@@ -71,6 +73,26 @@ export const generateVersionedModelId = <T, ModelIdKey extends keyof T>(
 export const incVersion = (version: NonNegativeInteger): NonNegativeInteger =>
   (Number(version) + 1) as NonNegativeInteger;
 
+const getPartitionKeyFromSearchKey = <
```
gunzip, line 76: What about adding a comment here?

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -71,6 +73,26 @@ export const generateVersionedModelId = <T, ModelIdKey extends keyof T>(
...
+  T,
+  ModelIdKey extends keyof T,
+  PartitionKey extends keyof T = ModelIdKey
+>(
+  searchKey: DocumentSearchKey<T, ModelIdKey, PartitionKey>
+): CosmosPartitionKey => {
+  const builder = new PartitionKeyBuilder();
```
gunzip, line 83: why do we need these changes? I read "If you are not using Hierarchical Partitioning feature, Definition of Partition Key is practically backward compatible" here:

https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/cosmosdb/cosmos/CHANGELOG.md

.yarn/releases/yarn-classic.cjs
balanza, line 1: **issue:** ‏Is it needed by the changes proposed in the PR? If there is a need (ex: _cannot import new module because reasons_) I advocate we highlight the change and its consequences in the PR description.

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -71,6 +73,26 @@ export const generateVersionedModelId = <T, ModelIdKey extends keyof T>(
...
+  const value: unknown =
+    searchKey[1] !== undefined ? searchKey[1] : searchKey[0];
+  if (
+    typeof value === "string" ||
+    typeof value === "number" ||
+    typeof value === "boolean"
+  ) {
+    builder.addValue(value);
+  }
+  return builder.build();
```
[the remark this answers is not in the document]
lucacavallaro: then the pk is "none"

src/utils/cosmosdb_model_versioned.ts
lucacavallaro: because our `partitionKey` has this type annotation:

```ts
> = (T & BaseModel)[ModelIdKey] extends string // narrow type to the ones that might be an identity
  ? PartitionKey extends ModelIdKey // eslint-disable-next-line functional/prefer-readonly-type
    ? [(T & BaseModel)[ModelIdKey]] // eslint-disable-next-line functional/prefer-readonly-type
    : PartitionKey extends keyof (T & BaseModel) // eslint-disable-next-line functional/prefer-readonly-type
    ? [(T & BaseModel)[ModelIdKey], (T & BaseModel)[PartitionKey]]
    : never
  : never;
```

that is not compatible with the `PartitionKey` type defined by `@azure/cosmos`

lucacavallaro: aaaaaaaaa! I understand now what you mean, I can remove the builder!

src/utils/cosmosdb_model_versioned.ts
```diff
@@ -71,6 +73,26 @@ export const generateVersionedModelId = <T, ModelIdKey extends keyof T>(
...
+const getPartitionKeyFromSearchKey = <
+  T,
+  ModelIdKey extends keyof T,
+  PartitionKey extends keyof T = ModelIdKey
+>(
+  searchKey: DocumentSearchKey<T, ModelIdKey, PartitionKey>
+): CosmosPartitionKey => {
+  const builder = new PartitionKeyBuilder();
+  const value: unknown =
+    searchKey[1] !== undefined ? searchKey[1] : searchKey[0];
```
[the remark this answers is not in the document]
lucacavallaro: I decided to make it more explicit using `typeof x !== "undefined"`

src/utils/cosmosdb_model_versioned.ts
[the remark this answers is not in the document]
balanza: I don't understand, I will need to read the docs for `PartitionKeyBuilder` (_which is a signal we might need an explanatory comment for occasional readers to grasp the needed concepts_).

From a high-level view, I see a hidden behavior: users will perform a query with a different value they passed.
I see we are doing it to narrow the type of `value` according to the expectations of `addValue`. I think we can either:
1. throw an exception on the else branch, as the case can never happen
2. refine the definition of `DocumentSearchKey` to accept only number, string and boolean (which I guess cover all the real-life scenarios)

What do you think?

.yarn/releases/yarn-classic.cjs
BurnedMarshal, line 1: Should be the `.yarn` folder added into the `.gitignore` file?

lucacavallaro: nope, that folder is part of yarn configuration

gunzip (approved):

src/utils/cosmosdb_model_versioned.ts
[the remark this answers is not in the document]
balanza: refactored in [8d1fb6f](https://github.com/pagopa/io-functions-commons/pull/371/commits/8d1fb6f34dfeb8e9373b7232634cc4a7c38355ea)

balanza (approved):

package.json
```diff
@@ -102,5 +102,6 @@
   },
   "resolutions": {
     "graceful-fs": "^4.2.4"
-  }
+  },
+  "packageManager": "yarn@4.1.1"
```
balanza, line 106: **question:** I'm a bit confused about this. We are adding `.yarn/releases/yarn-classic.cjs` file so that newer version of yarn won't try to update to v4, then we set v4 as default package manager. Am I missing something?‏

lucacavallaro: `packageManager` tells `corepack` to load `yarn@4.1.1` that then loads `yarn classic` according to the `yarnrc.yml` configuration

the commit files was generated by running the command `yarn set version classic`, and according to the [yarn set version documentation](https://yarnpkg.com/cli/set/version)

> This command will set a specific release of Yarn to be used by Corepack: https://nodejs.org/api/corepack.html.

> By default it only will set the [packageManager](https://yarnpkg.com/configuration/manifest#packageManager)  field at the root of your project, but if the referenced release cannot be represented this way, if you already have [yarnPath](https://yarnpkg.com/configuration/yarnrc#yarnPath) configured, or if you set the --yarn-path command line flag, then the release will also be downloaded from the Yarn GitHub repository, stored inside your project, and referenced via the [yarnPath](https://yarnpkg.com/configuration/yarnrc#yarnPath) settings from your project .yarnrc.yml file.

I think that `classic` is not a valid `packageManager` value for `corepack` - maybe tomorrow I can try to set `1.22`

gunzip (approved):

BurnedMarshal (approved):

gunzip (approved):

PR #372, opened 2024-03-28
[DEVEX-106] Fix proposal

balanza:
I propose a fix on #371 

The `DocumentSearchKey` type ensures we query database object by fields whose type is `string` (for model id) or `string | number` (for partition key)

### DIFF
--- a/src/utils/cosmosdb_model.ts
+++ b/src/utils/cosmosdb_model.ts
@@ -57,13 +57,19 @@ export type DocumentSearchKey<
   // Hence we omit "extends BaseModel", but we check keys to be part of "T & BaseModel"
   ModelIdKey extends keyof (T & BaseModel),
   PartitionKey extends keyof (T & BaseModel) = ModelIdKey
-> = (T & BaseModel)[ModelIdKey] extends string // narrow type to the ones that might be an identity
-  ? PartitionKey extends ModelIdKey // eslint-disable-next-line functional/prefer-readonly-type
-    ? [(T & BaseModel)[ModelIdKey]] // eslint-disable-next-line functional/prefer-readonly-type
-    : PartitionKey extends keyof (T & BaseModel) // eslint-disable-next-line functional/prefer-readonly-type
-    ? [(T & BaseModel)[ModelIdKey], (T & BaseModel)[PartitionKey]]
-    : never
-  : never;
+> =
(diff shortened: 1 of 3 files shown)


src/utils/cosmosdb_model.ts
```diff
@@ -57,13 +57,19 @@ export type DocumentSearchKey<
...
+  // the modelId must be a string
+  // the partitiion key might be a string or a number
+  Pick<BaseModel & T, ModelIdKey | PartitionKey> extends Record<
+    ModelIdKey,
+    string
+  > &
+    Record<PartitionKey, string | number>
+    ? PartitionKey extends ModelIdKey // partition key === model id means no partition key is provided
+      ? readonly [string]
+      : readonly [string, string | number]
```
lucacavallaro, line 71: maybe we should also support boolean?

```suggestion
      : readonly [string, string | number | boolean]
```

lucacavallaro (approved):

src/utils/cosmosdb_model.ts
```diff
@@ -57,13 +57,19 @@ export type DocumentSearchKey<
...
+> =
+  // We must be sure the provided keys refer to fields whose value is appropriate
+  // the modelId must be a string
+  // the partitiion key might be a string or a number
+  Pick<BaseModel & T, ModelIdKey | PartitionKey> extends Record<
+    ModelIdKey,
+    string
+  > &
+    Record<PartitionKey, string | number>
+    ? PartitionKey extends ModelIdKey // partition key === model id means no partition key is provided
```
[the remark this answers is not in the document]
balanza: Being both `PartitionKey` and `ModelIdKey` literals, `PartitionKey extends ModelIdKey` means they are the same.
This happens in the scenario in which no partition key is provided
```ts
DocumentSearchKey<MyModel, "id"> //  ModelIdKey="id" and PartitionKey="id"
```
We field type is checked above

src/utils/cosmosdb_model.ts
balanza: Technically yes. What boolean partition key you imagine to use?

lucacavallaro: I can't imagine any! I was just wondering to "adhere" to the `PrimaryKey` type of `@azure/cosmos` which puts it among the supported scalars

PR #375, opened 2024-05-09
[IOCOM-1362] Added temporary model

michaeldisaro:
#### List of Changes
Added a temporary non versioned model with some defaults to get rc configurations as if they are non versioned.

#### Motivation and Context
We need to migrate to a non versioned model.

#### Types of changes

### DIFF
--- a/src/models/rc_configuration_non_versioned_temp.ts
+++ b/src/models/rc_configuration_non_versioned_temp.ts
@@ -1,6 +1,5 @@
 import * as t from "io-ts";
 import * as TE from "fp-ts/lib/TaskEither";
-import * as RA from "fp-ts/lib/ReadonlyArray";
 import * as O from "fp-ts/lib/Option";
 
 import {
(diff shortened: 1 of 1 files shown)


src/models/rc_configuration_non_versioned_temp.ts
```diff
@@ -0,0 +1,137 @@
...
+    "hasPrecondition"
+  ),
+  isLollipopEnabled: t.boolean,
+  name: NonEmptyString,
+  userId: NonEmptyString
+});
+
+const RCConfigurationO = t.partial({
+  version: NonNegativeInteger
+});
```
Garma00, line 74: Do we need this?

src/models/rc_configuration_non_versioned_temp.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export const RCConfigurationBase = t.intersection([
+  RCConfigurationR,
+  RCConfigurationO
```
Garma00, line 78: Ditto

Garma00 (approved):
I was thinking if we do not need the versioned one any longer what about using this model in the unit test already defined for the versioned one (with fix if needed)?

src/models/rc_configuration_non_versioned_temp.ts
michaeldisaro: yep.

src/models/rc_configuration_non_versioned_temp.ts
michaeldisaro: yep

acuiuli (approved):

PR #382, opened 2024-06-10
[IOCOM-1504] Added new class method to get blob from most recent ContainerClient SDK

michaeldisaro:
#### List of Changes
Added new class method to get blob from most recent ContainerClient SDK.
Added test.

#### Motivation and Context
We need to move to most recent SDK because `azure-storage` has been deprecated and will not support identity  based authentication.
The new `@azure/blob-storage` SDK contains the new `ContainerClient` that allows to use identity based connection.

Unit Test

#### Types of changes

### DIFF
--- a/src/models/message.ts
+++ b/src/models/message.ts
@@ -33,6 +33,7 @@ import { TimeToLiveSeconds } from "../../generated/definitions/TimeToLiveSeconds
 import {
   GenericCode,
   getBlobAsTextWithError,
+  getBlobFromContainerAsTextWithError,
   upsertBlobFromObject
 } from "../utils/azure_storage";
 import { wrapWithKind } from "../utils/types";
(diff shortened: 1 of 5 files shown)


src/models/message.ts
```diff
@@ -449,4 +451,65 @@ export class MessageModel extends CosmosdbModelTTL<
...
+        TE.fromEither(
+          E.fromOption(
+            // Blob exists but the content is empty
+            () => ({
+              code: GenericCode,
+              message: "Cannot get stored message content from empty blob"
+            })
+          )(maybeContentAsText)
+        )
+      ),
```
Garma00, line 483: I think we can simplify with

```suggestion
      TE.chain(
        TE.fromOption(
          // Blob exists but the content is empty
          () => ({
            code: GenericCode,
            message: "Cannot get stored message content from empty blob"
          })
        )
      ),
```

michaeldisaro: This code is intended to be "identical" to the other one, at least on a first step. We need it to behave as much identically as possibile.

PR #384, opened 2024-11-15
[#IOPID-2428] Update `ValidationTokenEntity` to support `@azure/data-tables`

salvatorelaiso:
#### List of Changes

- Create `TableEntityAzureDataTables` to handle the new camel case `partitionKey` and `rowKey` of `@azure/data-tables` library.
- Create `ValidationTokenEntityAzureDataTables` to use the new `TableEntityAzureDataTables`

#### Motivation and Context

The library `@azure/data-tables` returns an Entity with the partition and row keys in camel case, opposed to the pascal case used by the deprecated library `azure-storage`.
Since the functions  such `io-functions-public` as are being updated with the newest library, the validation of the entities will fail on the mismatched keys.
This PR will address this issue by exposing a new `TableEntityAzureDataTables` and `ValidationTokenEntityAzureDataTables` which can be used to validate an Entity retrieved with `@azure/data-tables`.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/entities/validation_token.ts
+++ b/src/entities/validation_token.ts
@@ -7,8 +7,8 @@ import * as t from "io-ts";
 import { UTCISODateFromString } from "@pagopa/ts-commons/lib/dates";
 import { EmailAddress } from "../../generated/definitions/EmailAddress";
 import { FiscalCode } from "../../generated/definitions/FiscalCode";
-import { TableEntityAzureStorage } from "../utils/azure_storage";
-import { TableEntity } from "../utils/data_tables";
+import { TableEntity } from "../utils/azure_storage";
+import { TableEntityAzureDataTables } from "../utils/data_tables";
 
 export const VALIDATION_TOKEN_TABLE_NAME = "ValidationTokens";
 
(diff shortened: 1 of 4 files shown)


src/utils/azure_storage.ts
```diff
@@ -263,15 +263,17 @@ export const getBlobAsObject = async <A, O, I>(
 
 // TABLE STORAGE FUNCTIONS AND TYPES
 
-// Basic type for a table entity
+/**
+ * Legacy. Basic type for a table entity for the library `azure-storage`.
+ */
 /* eslint-disable @typescript-eslint/naming-convention */
-export const TableEntity = t.interface({
+export const TableEntityAzureStorage = t.interface({
```
arcogabbo, line 270: since we are changing an already exported and well-known type `TableEntity`, this type of change will introduce a breaking change at runtime for old adopters. I would instead leave the name as it is and create a new type called for example `TableEntityAzureDataTables` inside `utils/data_tables.ts` for new adopters, eliminating a breaking change. do you agree?

src/utils/data_tables.ts
```diff
@@ -0,0 +1,8 @@
+import * as t from "io-ts";
+
+export const TableEntity = t.type({
```
arcogabbo, line 3: see above comment

arcogabbo (changes requested):

src/utils/azure_storage.ts
salvatorelaiso: You're right. I'll change it right away.

salvatorelaiso: Resolved in [491818b](https://github.com/pagopa/io-functions-commons/pull/384/commits/491818b5019eaf91be790915b87b44909b57785e)

src/utils/data_tables.ts
salvatorelaiso: Resolved in [491818b](https://github.com/pagopa/io-functions-commons/pull/384/commits/491818b5019eaf91be790915b87b44909b57785e)

src/entities/validation_token.ts
```diff
@@ -21,9 +22,24 @@ export const ValidationToken = t.interface({
...
 
+export const ValidationTokenEntityAzureDataTables = t.intersection([
+  ValidationToken,
+  TableEntityAzureDataTables
+]);
+
 export const ValidationTokenEntity = t.intersection([
   ValidationToken,
   TableEntity
 ]);
```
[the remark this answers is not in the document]
salvatorelaiso: If that's the case, then we can keep the upgraded one only and discard the version using the old `TableEntity`. Do you agree?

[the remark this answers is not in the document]
salvatorelaiso: See [4c33ce1](https://github.com/pagopa/io-functions-commons/pull/384/commits/4c33ce14f25cacd25ca6ac92722238aa3ff63067)

arcogabbo (changes requested):
please address errors detected by CI

arcogabbo (approved):

PR #429, opened 2026-07-08
[#IOPLT-1951] Add transport for OneMail

gquadrati:
#### List of Changes

WIP

#### Types of changes

#### Checklist:

- [ ] My change requires a changeset, which I've added.

### DIFF
--- a/src/mailer/onemail.ts
+++ b/src/mailer/onemail.ts
@@ -1,0 +1,344 @@
+/**
+ * Implements a Nodemailer OneMail transport.
+ *
+ * Uses the OneMail Dispatcher REST API to send transactional emails through the
+ * high-priority endpoint (POST /v1/emails/send/high) with direct HTML content:
+ * see https://github.com/pagopa/onemail
+ *
+ * Note: the high-priority endpoint accepts a single recipient and does not
+ * support attachments. Responses are asynchronous (202 Accepted): the API
+ * returns a requestId that can later be used to check the delivery status.
+ */
+import { EmailString, NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+import { readableReport } from "@pagopa/ts-commons/lib/reporters";
+import { withoutUndefinedValues } from "@pagopa/ts-commons/lib/types";
+import { isLeft, isRight } from "fp-ts/lib/Either";
+import * as E from "fp-ts/lib/Either";
+import {
+  fromEither,
+  fromPredicate,
+  TaskEither,
+  tryCatch
+} from "fp-ts/lib/TaskEither";
+import * as TE from "fp-ts/lib/TaskEither";
+import { fromNullable, Option } from "fp-ts/lib/Option";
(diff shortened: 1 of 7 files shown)


src/mailer/onemail.ts
```diff
@@ -0,0 +1,344 @@
...
+/**
+ * Translates a nodemailer parsed address ({ name: <name>, address: <address> })
+ * to the format expected by the OneMail API ({ name: <name>, email: <address> }),
+ * then get the first one from the input array.
+ */
+const toOneMailAddress = (
+  addresses: ReadonlyArray<NodemailerAddress>
+): Option<OneMailAddress> =>
+  pipe(
+    fromNullable(addresses[0]),
```
BurnedMarshal, line 196: A list of addresses is converted to only one inbox address silently and without any validation. To avoid mistakes we should enforce or a strict validation for the address in this trasporter or implements a multi-send command internally accepting a list of emails.
By design seems that the transactional send for OneMail (`high` API) support only one recipient. Multiple address delivery should custom implemented in here.

src/mailer/onemail.ts
```diff
@@ -0,0 +1,344 @@
...
+      // as they are just strings that can contain invalid addresses.
+      // Instead, mail.message.getAddresses() gets the email addresses
+      // already validated by nodemailer (or undefined in case there are
+      // no valid addresses for one specific field).
+      // The following cast exists because of a bug in nodemailer typings
+      // (MimeNode.Addresses are *not* just array of strings)
+      const addresses: IAddresses = mail.message.getAddresses() as IAddresses;
+
+      // The high-priority endpoint accepts a single recipient only.
+      if (addresses.to !== undefined && addresses.to.length > 1) {
```
BurnedMarshal, line 277: we need a similar validation even for the other addresses?
- only one from
- zero cc
- zero bcc 
- zero reply-to

src/mailer/onemail.ts
```diff
@@ -0,0 +1,344 @@
...
+        }),
+        extendedHeaders,
+        from: pipe(
+          fromNullable(addresses.from),
+          O.chain(toOneMailAddress),
+          O.toUndefined
+        ),
+        to: pipe(
+          fromNullable(addresses.to),
+          O.chain(toOneMailAddress),
```
BurnedMarshal, line 312: If we don't want handle a multi send mechanism (https://github.com/pagopa/io-functions-commons/pull/429/changes#r3549702763), maybe is better move the filter logic in here, instead to hide into the `toOneMailAddress` method.

```suggestion
          fromNullable(addresses.from[0]),
          O.chain(toOneMailAddress),
          O.toUndefined
        ),
        to: pipe(
          fromNullable(addresses.to[0]),
          O.chain(toOneMailAddress),
```
with 
```typescript
const toOneMailAddress = (
  addresses: NodemailerAddress
): Option<OneMailAddress> =>
```

